PR #4, opened 2021-05-24
Add ClusterNetworkPolicy User Stories -- Zero Trust Cluster

Dyanngg:


srampal (commented):
This should probably be reworded to state that all end user application pods cant communicate with each other by default. Stated as is, even the Kubernetes control plane components wont be able to communicate with each other. Then there has to be the minimum communication available for cluster add-on services such as CNIs, CSIs, DNS etc. So this use case needs to be specifically about user application level pods excluding cluster control plane pods and all add-on pods. This then further implies there should be a way to identify what are control plane and add-on pods in a common way. This may require that cluster admins follow a practice of specific annotations to identify control plane and add-ons pods and namespaces.

docs/userstories/ClusterNetworkPolicy/Zero_Trust_Cluster.md
```diff
@@ -0,0 +1,20 @@
+### Summary
+
+As a cluster admin, I want all Pods in my cluster to have a default zero-trust
+deny-all security posture, so that all traffic allowed in the cluster must be
+explicitly called out (by separate higher precedence policies).
```
abhiraut, line 5: an example of such exceptions would provide better context, as was brought up by Sanjeev in the meeting (eg open traffic to prometheus etc)

Dyanngg: Done

abhiraut:
tl;dr: this use case can be solved by DNP(default-deny-cluster) + CNP(allow critical traffic) + K8s Netpol(allow dev services)

This user story sounds like a justification for DefaultNetworkPolicy (or at least include some way in the API to write rules which can be overridden by devs)

So in order to achieve this user story, an admin must write default-deny policy for the cluster (say using DNP or like) and complement it with higher order CNP rules to allow cluster critical traffic (and traffic necessary for devs to operate). Devs are now forced to open up traffic for their own services using K8s Netpol and thereby disallowing unintentional allow of traffic.

PR #7, opened 2021-05-24
Add ClusterNetworkPolicy User Stories -- Funnel Cluster Traffic

Dyanngg:


astoycos:
I think the name of this policy needs to moved away from `funnel` I know you specified 

```
It is worth noting that the Cluster-scoped NetworkPolicy APIs will not redirect
traffic, rather it can ensure that no traffic is allowed in/out except traffic
via the gateways.
```

However network policy will never "funnel" traffic maybe just "Ensure gress gateways" or something [along those lines](https://github.com/kubernetes/enhancements/pull/2522/files#r601672091).

docs/userstories/ClusterNetworkPolicy/Funnel_Cluster_Traffic.md
```diff
@@ -0,0 +1,22 @@
+### Summary
```
astoycos, line 1: In general I think this story may be impossible, and or impossible to implement from a CNI's perspective.  

For example what would be done in the case a developer creates an external facing service (Nodeport, externalIP, ect.) backed by host networked pods? 

Here traffic could gress the cluster freely, and since CNI's can't _see_ host networked pods network policy would have no effect.

Dyanngg:
> I think the name of this policy needs to moved away from `funnel` I know you specified
> 
> ```
> It is worth noting that the Cluster-scoped NetworkPolicy APIs will not redirect
> traffic, rather it can ensure that no traffic is allowed in/out except traffic
> via the gateways.
> ```
> 
> However network policy will never "funnel" traffic maybe just "Ensure gress gateways" or something [along those lines](https://github.com/kubernetes/enhancements/pull/2522/files#r601672091).

Re-worded

astoycos:
nit: I would change the name of the PR itself too :)

docs/userstories/ClusterNetworkPolicy/Funnel_Cluster_Traffic.md
astoycos: The solution for my concern is actually somewhat simple, we should just should clarify something like this for every ingress/egress related story : 

```
An administrator should be able to allow all (non-privileged, non-hostNetwork) 
pods in a cluster to send/receive external traffic  only via specified gateways.
```

PR #8, opened 2021-05-24
Add ClusterNetworkPolicy User Stories -- Cluster Compliance

Dyanngg:


srampal (commented):
Imo this is worded in a generic manner and should provide slightly more specific examples. The user story should indicate firstly that the cluster admin must be able to mandate cluster compliance, not just provide options for monitoring and cluster DNS. Some of the other user stories such as egress control or namespace isolation are also about enforcing cluster compliance so this story should have more specific details that are not covered by the other user stories.

docs/userstories/ClusterNetworkPolicy/Cluster_Compliance.md
```diff
@@ -0,0 +1,12 @@
...
+As a cluster admin, I want all workloads to start with a network/security
+model that meets the needs of my company.
+
+### Description
+
+A platform admin may want to factor out policies that each namespace would have
+to write individually in order to make deployment and auditability easier.
+Common examples include allowing all workloads to be able to talk to the cluster
+DNS service and, similarly, allowing all workloads to talk to the logging/monitoring
+pods running on the cluster.
```
astoycos, line 12: Please Include this conversation from the KEP as it's pretty relevant I think. 
https://github.com/kubernetes/enhancements/pull/2522/files#r601709104

astoycos, line 12: In general I think this story needs to go into detail and provide examples regarding the actual expected behavior here, i.e _who_ (if anyone) can override these baseline polices and _how_ can they do it.

abhiraut: agreed .. we need to explicitly call out whether the expectation here is that these policies are baseline? or something more strict. since its compliance, it must be strict (hence not a justification of DNPs)

PR #14, opened 2021-06-08
Add more detailed instructions about user stories and new Objects 

astoycos:
Add some more notes on how to form and organize new userstories, following the existing k8's readme formatting rules 

Create two empty directories for `v1` and `v2` apis


abhiraut (commented):
/lgtm

PR #29, opened 2022-02-25
[WIP] Import cyclonus to network-policy-api repo 

astoycos:
This PR moves https://github.com/mattfenwick/cyclonus to this repo, with all the legacy commits and tags.


mattfenwick:
What's next steps for this @astoycos ?

astoycos:
So really there's nothing blocking this @mattfenwick , however I do think It would easier to let https://github.com/kubernetes-sigs/network-policy-api/pull/30 merge first to make sure we work around the directory structure/tools of the CRD since it's a bit more crucial we get that in.

astoycos:
Close in favor of #39

PR #30, opened 2022-02-28
Implement the AdminNetworkPolicy api CRD 

astoycos:
This PR serves to implement the AdminNetworkPolicy CRD described in further detail with https://github.com/kubernetes/enhancements/issues/2091. 

The goal is to ensure this does not merge without solving all of the TODO's described in the KEP.

PR #31, opened 2022-03-04
Setup API base for Implementation of AdminNetworkPolicyAPI

astoycos:
Setup the base directory structure and autogenerated kubebuilder files to generate the new AdminNetworkPolicyAPI, Implementation of the API specifically will take place in a follow-up PR

https://book.kubebuilder.io/cronjob-tutorial/new-api.html


srampal:
1) Thanks for eliminating all the vendor'ed/ module dependencies .. as discussed offline .. it didnt seem right to commit 2695 files to define a reference CRD spec even if all the other files were downloaded go module dependencies.
2) For the remaining change, set, I'd seek guidance from sig-network/ @thockin on what exactly we should commit for now in order to define the reference CRD .. trying to understand the process of which artifacts need to be committed by a working group defining a v1alpha1 CRD.. should it be an entire set of kubebuilder scaffolding ? or the *_types.go ? or something else ? Also, CRDs are sometimes versioned v1 even if they are at the "alpha1 "stage .. maybe others/ @thockin can clarify that as well.

controllers/adminnetworkpolicy_controller.go
```diff
@@ -0,0 +1,63 @@
+/*
```
astoycos, line 1: For reviewers: We shouldn't need anything in `controllers/` for now since the implementation of the ANP API will occur in CNIs so either I can remove it now or in a followup PR

danwinship:
> it didnt seem right to commit 2695 files to define a reference CRD spec even if all the other files were downloaded go module dependencies.

If the vendored files are used even as part of the build process then they should be committed. Otherwise different people will end up getting different versions of them and then getting different autogenerated results

astoycos:
Thanks @danwinship! I think in this case since we won't be shipping any controller/binaries related to the ANP API we'll leave the vendor files out of git tracking here.... The only time a process is run _should_ be when a user updates the API and the manifests need to be re-generated, which is handled by the controller-gen binary, stored within this repository 

Since this PR is now much smaller I am going ahead and just leave it all (the API definition + kubebuilder auto generated files) in a single PR 

Closing this in favor of https://github.com/kubernetes-sigs/network-policy-api/pull/30

astoycos:
@srampal ^^

PR #41, opened 2022-09-16
Bump python Version and change name

astoycos:
Fixes due to a build error and the website url `kubernetes-sigs-adminnetworkpolicy-api` being too long

See https://github.com/kubernetes/org/issues/3643#issuecomment-1248234328 for more details


tssurya:
```
4:20:45 PM: $ make docs
4:20:45 PM: ./hack/make-docs.sh
4:20:45 PM: ./hack/make-docs.sh: line 36: mkdocs: command not found
4:20:45 PM: make: *** [Makefile:49: docs] Error 127
```
Is this something we need to fix before merging this?

tssurya:
/lgtm
thanks Andrew

PR #58, opened 2022-11-08
Add Website and Other Links to README

cailynse:
As recommended by @astoycos I've added some additional content to the README.

Questions:
Some other READMEs have nice graphics - is there one that would make sense here?
As a new contributor I find the website being only for the AdminNetworkPolicy to be a bit confusing - is it worth either specifying that it is the current team focus - or adding a more generic landing page to the website?
Should we link to any NW Policies v1 site/documentation incase someone ends up here by mistake?


astoycos:
/assign

Thanks so much for this @cailynse!! Will take a look ASAP

astoycos (commented):
> Some other READMEs have nice graphics - is there one that would make sense here?

Yeah I think for starters we can just put a K8s logo? Eventually it'd be super cool to have our own :) 

> As a new contributor I find the website being only for the AdminNetworkPolicy to be a bit confusing - is it worth either specifying that it is the current team focus - or adding a more generic landing page to the website?

100 % we should definitely do that I made an issue to track it. https://github.com/kubernetes-sigs/network-policy-api/issues/59

> Should we link to any NW Policies v1 site/documentation incase someone ends up here by mistake?

Yep good idea!

astoycos:
Also small nit @cailynse But can you sign your commits with `-s`? It just made it easier to figure out who owned the original commits in the future!

public.pgp
```diff
@@ -0,0 +1,67 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
astoycos, line 1: @cailynse I don't _think_ we need this :)

astoycos: No Worries!!

astoycos:
Thanks @cailynse This looks sooo much better!

PR #60, opened 2022-11-15
Update cyclonus makefile to figure out cpu arch

olivercodes:
@mattfenwick @astoycos - I realize you might want to take a different approach with this (such as the tag), so no worries if you want to post back with other modifications we should add to cyclonus build/publish. I updated it whilst working on the ANP conformance tests setup (on my m1).

### DIFF
--- a/cmd/cyclonus/Makefile
+++ b/cmd/cyclonus/Makefile
@@ -1,3 +1,6 @@
+HOST_GOOS ?= $(shell go env GOOS)
+HOST_GOARCH ?= $(shell go env GOARCH)
+
 test:
 	go test ./pkg/...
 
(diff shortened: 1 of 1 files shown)


mattfenwick:
hi @olivercodes thanks for putting this together !  

_( btw even though I'm listed as an approver ? , I shouldn't have final say on what does and doesn't go into this repo now that it's been upstreamed cc @astoycos , should somebody else get added to [reviewers](https://github.com/kubernetes-sigs/network-policy-api/blob/master/cmd/cyclonus/OWNERS) ? )_

it's been a while and I'm not even sure if the makefile was used , check out https://github.com/kubernetes-sigs/network-policy-api/blob/master/cmd/cyclonus/.goreleaser.yml which I used for releases ... _also please note that lots of things haven't been changed yet -- things that can't work in this repo -- like pushing images to my personal dockerhub account :smile:_

 - spin off issue: how should builds be handled long term , cc @astoycos 

I don't 100% understand the goal of this change, if you wouldn't mind dropping a couple lines of motivation/goal it'll help me to understand what/what/how !

olivercodes:
@mattfenwick:

Totally understandable. I was actually hoping to get more information like what you just posted! So this is a good start.

I had 2 goals with the change:
1) get a short convo going around long-term support/maintenance for cyclonus itself. I was also thinking to update the Cyclonus build process (for long-term needs as you mentioned) whilst updating and working on the conformance tests for ANP, but didn't want to do anything there without some guidance on requirements in doing so.
2) If it's useful, updating the makefile for the (probably very small number of) people that want do do local builds on a different architecture.

On #2, I think we could also just delete the file, and document a basic build command in the developer setup readme (as you have done here): https://github.com/kubernetes-sigs/network-policy-api/blob/master/cmd/cyclonus/docs/developer-guide.md#setup

olivercodes:
Closing in favor of #62

PR #80, opened 2023-03-13
Fix codegen

astoycos:
In our api definition we were missing the
`// +genclient:nonNamespaced`
tag which resulted in our generated listers being incorrectly namespace
based.

I added this tag and went to regenerate the informers,clients, and listers
but things weren't working correctly.

This was because the code wasn't stored at `$GOPATH sigs.k8s.io/network-policy-api`.

Add a step to the makefile to make sure the generate targets always work
regardless of where the code lives on local disk.

Also move the generated clients to `pkg` to keep up to date with standard
go best practices: https://github.com/golang-standards/project-layout

Lastly fix the verify-codegen script to work with the new changes.


tssurya (approved):
/lgtm
LGTM,

good call on moving the client stuff into pkg, I see gateway-api has the same structure.

PR #91, opened 2023-05-05
Add more docs, refactor website, add release tooling

### LINKED ISSUE #59: Refactor website
As part of this repo we've created the following website: https://network-policy-api.sigs.k8s.io/

Currently it takes you straight to the AdminNetworkPolicy API information however the network-policy-api repo was designed to be broader than that and instead house extra information and tools for all past and present policy API's 

Therefore we should refactor the website to follow suite.  Specifically we should 

- [ ] Make a new landing page describing the subgroup and what we do -> #91 
- [ ] Add a section for APIs 
- [ ]Add a section for Tooling 
    - [ ]Add a bit about Cyclonus and why we have it  
- [ ]Add a section for In progress Initiatives and design documents 

along with anything else we may see fit :)

### LINKED ISSUE #89: Add Release tooling and Documentation
This issue will track the addition of release tooling and documentation.  Much of it can be carried over from the gateway-api project, except for now we won't need to handle release channels since we're just getting started.

This will allow us to finish #90

astoycos:
Partially fixes #59 
Fixes #89


astoycos:
Please take a look at these changes + the new version of the website in the netlify preview and let me know if it needs to change at all :) 

Getting this in _should_ allow us to cut a release of v1alpha1 early next week.

astoycos:
/hold 

so that it doesn't automatically merge with an lgtm

config/crd/policy.networking.k8s.io_baselineadminnetworkpolicies.yaml
```diff
@@ -1,10 +1,9 @@
----
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/2522
-    controller-gen.kubebuilder.io/version: v0.11.1
+    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/1923
```
astoycos, line 5: Wrong one

astoycos:
Fixed Up following the conversation in our sig-network-policy-api meeting today

hack/update-codegen.sh
```diff
@@ -0,0 +1,94 @@
...
+  go run k8s.io/code-generator/cmd/register-gen \
+    --input-dirs "${APIS_PKG}/apis/${VERSION}" \
+    --output-package "${APIS_PKG}/apis/${VERSION}" \
+    ${COMMON_FLAGS}
+
+  echo "Generating ${VERSION} deepcopy at ${APIS_PKG}/apis/${VERSION}"
+  go run sigs.k8s.io/controller-tools/cmd/controller-gen \
+    object:headerFile=${SCRIPT_ROOT}/hack/boilerplate.generatego.txt \
+    paths="${APIS_PKG}/apis/${VERSION}" 
+done
```
Dyanngg, line 94: nit: I think we need newline here

astoycos: Done

site-src/api-overview.md
```diff
@@ -1,13 +1,16 @@
...
+Application Developer, although in reality it is used by many different cluster
+personas, sometimes creating a complex web of objects to be maintained. In
+Contrast, each resource in the Network Policy API is designed to be used by a
+specific persona.
 
-With the advent of the AdminNetworkPolicy API Cluster Admins will now have the 
-ability to apply policy on in-cluster workloads with only a few simple policy 
-objects that can be applied globally.
+With the advent of the AdminNetworkPolicy and BaselineAdminNetworkPolicy
+resources Cluster Admins will now have the ability to apply policy on in-cluster
```
tssurya, line 12: nit: I know the first version only supports `in-cluster workloads` but in this part I'd keep it more generic since we mention this as a limitation already under General notes.

site-src/implementations.md
```diff
@@ -0,0 +1,16 @@
...
+Network Policy API resources and provides status and resource references for them.
+
+Implementors of Network Policy API resources are encouraged to update this document with status information about their
+implementations, the versions they cover, and documentation to help users get started.
+
+## AdminNetworkPolicy and BaseLineAdminNetworkPolicy
+
+Updated: 05-08-2023
+
+- [Ovn-Kubernetes CNI](https://github.com/ovn-org/ovn-kubernetes/pull/3489) (work in progress)
```
tssurya, line 13: nit: `OVN-Kubernetes CNI`

site-src/implementations.md
```diff
@@ -0,0 +1,16 @@
+# Implementations
+
+This document tracks downstream(actual and planned) implementations of
+Network Policy API resources and provides status and resource references for them.
+
+Implementors of Network Policy API resources are encouraged to update this document with status information about their
```
tssurya, line 6: ++

RELEASE.md
```diff
@@ -1,9 +1,95 @@
...
-The Kubernetes Template Project is released on an as-needed basis. The process is as follows:
+## Overview
 
-1. An issue is proposing a new release with a changelog since the last release
-1. All [OWNERS](OWNERS) must LGTM this release
-1. An OWNER runs `git tag -s $VERSION` and inserts the changelog and pushes the tag with `git push $VERSION`
-1. The release issue is closed
-1. An announcement email is sent to `kubernetes-dev@googlegroups.com` with the subject `[ANNOUNCE] kubernetes-template-project $VERSION is released`
+The Network Policy API project is has the following two main release components:
+- Kubernetes Custom Resource Definitions (CRDs)
```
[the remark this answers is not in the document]
tssurya: In fact I like what we have in the versioning page:
```
This contains the following:

API Types (Go bindings for the resources)
CRDs (Kubernetes definitions of the resources)
```
more clear?

site-src/versioning.md
```diff
@@ -0,0 +1,102 @@
...
+
+#### Beta -> GA
+
+A resource to graduate from beta to GA must meet the following criteria:
+
+* Almost all of the fields and behavior have conformance test coverage.
+* Multiple conformant implementations.
+* Widespread implementation and usage.
+* At least 6 months of soak time as a beta API.
+* Approval from subproject owners + KEP reviewers.
```
tssurya, line 77: We can add scale testing here.

pkg/generator/main.go
```diff
@@ -30,7 +30,7 @@ const (
 
 	// These values must be updated during the release process
 	bundleVersion = ""
-	approvalLink  = ""
+	approvalLink  = "https://github.com/kubernetes-sigs/network-policy-api/issues/90"
```
tssurya, line 33: nit this step is not documented anywhere, the fact that the api-approved annotation and this approval Link must be set to an issue number for release that needs to be opened...?
and we don't need link any for the bundle version?

config/crd/policy.networking.k8s.io_baselineadminnetworkpolicies.yaml
```diff
@@ -1,10 +1,9 @@
----
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/2522
-    controller-gen.kubebuilder.io/version: v0.11.1
+    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/network-policy-api/issues/90
+    policy.networking.k8s.io/bundle-version: ""
```
tssurya, line 6: shouldn't the bundle-version value be "0.1.0" ? 

(nvm) later saw https://github.com/kubernetes-sigs/network-policy-api/pull/93/commits/fac2c0ee8630bc15e59c7b3a6bdb9fc2fffa95bb

tssurya (changes requested):
looking good, some comments/questions/suggestions
I think we are missing two things:

- we probably want to include Dan's suggestion in the description part of same and notsame lables saying "this is subject in future, not stable" ?
- we don't need https://github.com/kubernetes-sigs/network-policy-api/blob/master/apis/v1alpha1/adminnetworkpolicy_types.go#L34 and https://github.com/kubernetes-sigs/network-policy-api/blob/645dbc2e6f75a36fe2b5e3c615bd8e574f14ea3f/apis/v1alpha1/baselineadminnetworkpolicy_types.go#L29 anymore?

which makes me curious how the yamls were generated... is the `pkg/generator/main.go` overwriting what we do from codegen? because I see this change in the yaml but we never changed the types.go file so... was the yaml file changes manually done?

RELEASE.md
```diff
@@ -1,9 +1,95 @@
...
+provides a good starting point. Once you're satisfied with the changelog, create
+a PR. This must go through the regular PR review process and get merged into the
+`main` branch. Approval of the PR indicates community consensus for a new
+release.
+
+### Release Steps
+
+The following steps must be done by one of the [network-policy API maintainers][network-policy-api-team]:
+
+For a **PATCH** release:
```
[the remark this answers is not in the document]
astoycos: > Is 0.1.0 a PATCH release or a MAJOR.MINOR release?

That would be a Minor release whole 0.1.1 would be a patch release.  I think we should include docs for both Patch and Major/Minor relases since If we do go for a v1apha2 API release I think it would be a "patch" release (Maybe I'm not 100% sure) 

> Also wondering if we should call out the reference to gateway-api here for giving credit/reference?

Added

site-src/api-overview.md
```diff
@@ -1,13 +1,16 @@
-## API overview
+# Api Object Overview
```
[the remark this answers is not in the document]
astoycos: Done

site-src/versioning.md
astoycos: Hrm Idk how to frame this do we want "upstream scale testing?" It's hard to say "All features are proven to be scalable"

pkg/generator/main.go
astoycos: This is in the release.md 

```
- Update `pkg/generator/main.go` with the new semver tag and any updates to the API review URL.
- Run the following command `BASE_REF=vmajor.minor.patch make generate` which
  will update generated docs and webhook with the correct version info. (Note
  that you can't test with these YAMLs yet as they contain references to
  elements which wont exist until the tag is cut and image is promoted to
```

astoycos: Yeah we don't have "opening the API Review Issue" in here though

RELEASE.md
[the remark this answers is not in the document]
astoycos: Agreed I tried to clear this up

astoycos:
> which makes me curious how the yamls were generated... is the pkg/generator/main.go overwriting what we do from codegen? because I see this change in the yaml but we never changed the types.go file so... was the yaml file changes manually done?

So now we don't use the cotroller-gen CLI to do this the generator pkg does it by calling into the controller-gen go package.  

```
		log.Printf("generating CRD for %v\n", groupKind)

		parser.NeedCRDFor(groupKind, nil)
		crdRaw := parser.CustomResourceDefinitions[groupKind]

		// Inline version of "addAttribution(&crdRaw)" ...
		if crdRaw.ObjectMeta.Annotations == nil {
			crdRaw.ObjectMeta.Annotations = map[string]string{}
		}
		crdRaw.ObjectMeta.Annotations[bundleVersionAnnotation] = bundleVersion
		crdRaw.ObjectMeta.Annotations[apiext.KubeAPIApprovedAnnotation] = approvalLink

		// Prevent the top level metadata for the CRD to be generated regardless of the intention in the arguments
		crd.FixTopLevelMetadata(crdRaw)

		conv, err := crd.AsVersion(crdRaw, apiext.SchemeGroupVersion)
		if err != nil {
			log.Fatalf("failed to convert CRD: %s", err)
		}

		out, err := yaml.Marshal(conv)
		if err != nil {
			log.Fatalf("failed to marshal CRD: %s", err)
		}

		fileName := fmt.Sprintf("config/crd/%s_%s.yaml", crdRaw.Spec.Group, crdRaw.Spec.Names.Plural)
		err = os.WriteFile(fileName, out, 0o600)
		if err != nil {
			log.Fatalf("failed to write CRD: %s", err)
		}
	}
```

Generates the CRDs with the `"sigs.k8s.io/controller-tools/pkg/crd"` package, adds the specified annotations and writes the CRD to disk

RELEASE.md
```diff
@@ -1,9 +1,95 @@
...
+
+This output will likely need to be reorganized and cleaned up a bit, but it
+provides a good starting point. Once you're satisfied with the changelog, create
+a PR. This must go through the regular PR review process and get merged into the
+`main` branch. Approval of the PR indicates community consensus for a new
+release.
+
+### Release Steps
+
+The following steps must be done by one of the [network-policy API maintainers][network-policy-api-team]:
```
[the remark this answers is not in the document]
astoycos: Yep this needs to be fixed

site-src/api-overview.md
astoycos: Changed to just "to globally apply policy"

RELEASE.md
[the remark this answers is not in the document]
astoycos: https://github.com/kubernetes/org/pull/4217

RELEASE.md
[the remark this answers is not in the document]
astoycos: https://github.com/kubernetes/org/pull/4217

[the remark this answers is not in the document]
tssurya: thanks andrew! comment can be resolved..

RELEASE.md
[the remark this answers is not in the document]
tssurya: thanks andrew! comment can be resolved..

RELEASE.md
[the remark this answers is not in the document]
tssurya: > > Is 0.1.0 a PATCH release or a MAJOR.MINOR release?
> 
> That would be a Minor release whole 0.1.1 would be a patch release. I think we should include docs for both Patch and Major/Minor relases since If we do go for a v1apha2 API release I think it would be a "patch" release (Maybe I'm not 100% sure)

yeah me neither which is why I was like maybe we add at that point when we want to cross that bridge.. At this point we are cutting 0.1.0 and that is patch so let's keep that only? Because I thought v1alpha2 would just be 0.1.1 or something... Since gateway-api is complex they have so many types of this, we should rethink this for our needs since we will do a per API thingy, I doubt we will use major.minor release? Let's keep the process as simple as possible, but I won't hold this PR because you have extra docs for stuff we aren't using yet.

> 
> > Also wondering if we should call out the reference to gateway-api here for giving credit/reference?
> 
> Added

added where? :D I am going blind...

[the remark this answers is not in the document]
tssurya: ah found it! https://github.com/kubernetes-sigs/network-policy-api/pull/91/commits/25c0f29a3a15230e64d13378811b625072ba53d0#diff-4e63587151032fdaf6fb3011e6594e8a3b294309a92d67d15654cc04b98d1757R1

site-src/versioning.md
tssurya: Ah hmm something like "ensuring API fields are scalable" is good, surely burden is on implementations, BUT I would like to reconsider things like supporting upto 1000 priorities and same/nodeSame etc if no one can use it due to it not scaling well across multiple implementations...

tssurya:
> Generates the CRDs with the `"sigs.k8s.io/controller-tools/pkg/crd"` package, adds the specified annotations and writes the CRD to disk

+1 way neater. Almost lgtm: I think we still need to add a warning comment for same/notSame labels before the cut:
```
we probably want to include Dan's suggestion in the description part of same and notsame lables saying "this is subject in future, not stable" ?
```

astoycos:
+1 @tssurya I think we can go ahead and merge this, then do a final look at the tooling + any other small comments before we complete #93

tssurya:
ack fine with me, let's merge this!

PR #97, opened 2023-05-15
remove unactive code owners

astoycos:
This commit removes some code owners who are no longer active in the sig-network-policy-api subgroup.

@abhiraut @rikatz PTAL just to make sure you are on board with this


rikatz:
/lgtm

I've been wanting to do this for some time, but forgot :)

Thanks Andrew!

PR #114, opened 2023-05-31
Adding a getting started section for installing the APIs

### LINKED ISSUE #76: Fix clean vendoring of repo for enabling plugin implementations
- [ ] document how it needs to be done
- [x] should we be vendoring in the `config` part of the repo as well? -> I'd say yes rather than letting each plugin generate their own yamls, we should be using the `kustomization.yaml` uniformly
- [x] If `kustomization.yaml` isn't used and plugins directly generate yamls it will break them at install time since the api-approved annotation will be missing.

These are some issues that we ran into when doing OVNK implementation for making the controller install CRD, so we should certainly clarify these.

I am waiting for the clientget changes to land first to try the vendoring again since there is some reordering of repos as a result of that.

Dyanngg:
closes #76

### DIFF
--- a/site-src/getting-started.md
+++ b/site-src/getting-started.md
@@ -1,0 +1,24 @@
+# Getting started with Network Policy API
+
+**1. Install a Network Policy API compatible CNI**
+
+There are numerous Container Network Plugin projects that support or are actively working on
+supporting the Network Policy API. Please refer to the [implementations](implementations.md)
+doc for details on the supportability matrix.
+
+**2. Install the Network Policy API CRDs**
+
+The following commands will install the latest release version of the `AdminNetworkPolicy` and
+`BaselineAdminNetworkPolicy` CRDs in your K8s cluster:
+
+```bash
+kubectl apply -f https://github.com/kubernetes-sigs/network-policy-api/releases/download/v0.1.0/install.yaml
+```
+
+**3. Try out one of the sample yamls for specific user stories**
+
+- [Deny traffic at a cluster level](reference/examples.md#sample-spec-for-story-1-deny-traffic-at-a-cluster-level)
+- [Allow traffic at a cluster level](reference/examples.md#sample-spec-for-story-2-allow-traffic-at-a-cluster-level)
+- [Explicitly Delegate traffic to existing K8s Network Policy](reference/examples.md#sample-spec-for-story-3-explicitly-delegate-traffic-to-existing-k8s-network-policy)
+- [Create and Isolate multiple tenants in a cluster](reference/examples.md#sample-spec-for-story-4-create-and-isolate-multiple-tenants-in-a-cluster)
+- [Cluster Wide Default Guardrails](reference/examples.md#sample-spec-for-story-5-cluster-wide-default-guardrails)
--- a/site-src/reference/examples.md
+++ b/site-src/reference/examples.md
@@ -65,7 +65,7 @@ spec:
               app: kube-dns
 ```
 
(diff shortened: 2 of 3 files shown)


site-src/getting-started.md
```diff
@@ -0,0 +1,25 @@
...
+
+**2. Install the Network Policy API CRDs**
+
+The following commands will install the latest version of the `AdminNetworkPolicy` and 
+`BaselineAdminNetworkPolicy` objects in your K8s cluster:
+
+```bash
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/network-policy-api/master/config/crd/policy.networking.k8s.io_adminnetworkpolicies.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/network-policy-api/master/config/crd/policy.networking.k8s.io_baselineadminnetworkpolicies.yaml
+```
```
tssurya, line 17: I would also add a note or warning on: Since the master branch can have activate changes, for best results point to the released version of the API that will be stable?
example:
https://raw.githubusercontent.com/kubernetes-sigs/network-policy-api/v0.1.0/config/crd/policy.networking.k8s.io_adminnetworkpolicies.yaml
instead of master, cause that could break users if they point to master

site-src/reference/examples.md
```diff
@@ -65,7 +65,7 @@ spec:
               app: kube-dns
 ```
 
-### Story 3: Explicitly Delegate traffic to existing K8s Network Policy
+### Sample Spec for Story 3: Explicitly Delegate traffic to existing K8s Network Policy
```
tssurya, line 68: not entirely related to this specific PR but, the examples here seem a bit wrong to me :dagger: 
the `pub-svc-delegate-example` in its egress peers section is using both `namespaces` and `pods` at the same time which is not allowed? https://github.com/kubernetes-sigs/network-policy-api/blob/7321d8509d9bf425c9c5b0338db3c0fb2816b4c5/apis/v1alpha1/shared_types.go#L102
If a user uses that it will fail, Perhaps we should fix them or open an issue and I can copy some from the conformance test yamls which are tested working samples (I've been running them on my cluster) into the website.

tssurya (changes requested):
Nice I like the `Guides` section! 
https://deploy-preview-114--kubernetes-sigs-network-policy-api.netlify.app/getting-started/
Some comments otherwise lgtm,
@Dyanngg could you point to closes issue https://github.com/kubernetes-sigs/network-policy-api/issues/76 so that when this PR merges that gets closed automatically?

site-src/getting-started.md
astoycos: Even simpler this should be 

`kubectl apply -f https://github.com/kubernetes-sigs/network-policy-api/releases/download/v0.1.0/install.yaml`

astoycos: And we need to add a note to the release doc to make sure that version number is updated everytime a release is cut.

site-src/reference/examples.md
astoycos: Good catch this is wrong 

It should just be `pods` with that namespace selector and podSelector

@Dyanngg Can you fixup the examples here?

site-src/getting-started.md
Dyanngg: Done

site-src/reference/examples.md
Dyanngg: Fixed

tssurya (approved):
/lgtm

PR #121, opened 2023-06-27
add Dan Winship as an owner

### LINKED ISSUE #118: Add Dan Winship as a network-policy-api approver
@danwinship Has been extremely active within this community since the very beginning when we first implemented the API following KEP completion.  We also recently went from 3 down to 2 maintainers and are feeling the strain of this transition. Due to this it makes perfect sense to have him become a maintainer here.  I'm opening this issue to allow any contributors to voice any concerns or give their +1's.

astoycos:
Fixes #118

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -3,3 +3,4 @@
 approvers:
   - astoycos
   - Dyanngg
+  - danwinship


aojea:
+1
😄

Issue #122, opened 2023-07-10
[ENHANCEMENT] Improve Tenancy API


npinaeva:
**Is your enhancement request related to a problem? Please describe.**
Existing [API](https://github.com/kubernetes-sigs/network-policy-api/blob/429a9e6ae89d411f89d5a16aba38a5d920c969ee/apis/v1alpha1/shared_types.go#L127-L143) implementing [tenancy use case](https://network-policy-api.sigs.k8s.io/user-stories/#story-4-create-and-isolate-multiple-tenants-in-a-cluster) allows a wide set of configs, some of them may be more confusing than useful. One example is the relationships between subject and tenancy https://docs.google.com/presentation/d/1TbWWaDgd2O-nXwNdlNmLXng1aQQ2i5cG/edit#slide=id.p7

**Describe the solution you'd like**
I would like to create a very simple API that covers required use cases, but is less powerful/confusing.

**Describe alternatives you've considered**
[A clear and concise description of any alternative solutions or features you've considered.](https://docs.google.com/document/d/113xBe7VMK7hMYdIdB9gobp7JwVkWQLnqdMPNkamfaK8/edit)

PR #123, opened 2023-07-11
Create NPEP-122 for Tenancy API update.

npinaeva:
Change Tenancy implementation from SameLabels/NotSameLabels ANP peers to a separate object
that will define Tenancy considering subject and peers are related.

Tracking issue https://github.com/kubernetes-sigs/network-policy-api/issues/122

### DIFF
--- a/npeps/npep-122.md
+++ b/npeps/npep-122.md
@@ -1,0 +1,191 @@
+# NPEP-122: Tenancy API
+
+* Issue: [#122](https://github.com/kubernetes-sigs/network-policy-api/issues/122)
+* Status: Provisional
+
+## TLDR
+
+Tenancy definition and the API (SameLabels/NotSameLabels) is confusing and ambiguous as of now. 
+We want to rethink the tenancy use cases and the API to keep it simple and easy to understand, but
+flexible enough to cover all defined use cases.
+
+## Goals
+
+- Clarify tenancy use cases.
+- Provide a complete definition of a tenant in respect to ANP.
+- Avoid unneeded tenancy configs that only exist because of ANP use cases.
+
+## Non-Goals
+
+- Define new use cases for tenancy (as opposed to clarify the ones we considered initially, but didn't explain well enough)
+- Define user stories for multiple tenancy policies in the same cluster.
+
+## Introduction
+
+The KEP doesn’t say a whole lot about tenants…
+
+From the Goals:
+
+    As a cluster administrator, I want to have the option to enforce in-cluster network level access controls that 
+    facilitate network multi-tenancy and strict network level isolation between multiple teams and tenants sharing 
+    a cluster via use of namespaces or groupings of namespaces per tenant.
+    
+    Example: I would like to define two tenants in my cluster, one composed of the pods in foo-ns-1 and foo-ns-2 
+    and the other with pods in bar-ns-1, where inter-tenant traffic is denied.
+
+From the User Stories:
+
+    Story 4: Create and Isolate multiple tenants in a cluster
+    
+    As a cluster admin, I want to build tenants in my cluster that are isolated from each other by default. 
+    Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
+    may own more than 1 Namespace.
+
+Elsewhere:
+
+    AdminNetworkPolicy Pass rules allows an admin to delegate security posture for certain traffic to the Namespace 
+    owners by overriding any lower precedence Allow or Deny rules. For example, intra-tenant traffic management can be 
+    delegated to tenant admins explicitly with the use of Pass rules.
+    
+So really, the only solidly-agreed-upon use case is that you should be able to create enforced isolation 
+between particular sets of namespaces.
+
+### Clarifying user stories
+
+#### Story 4.1: Create and Isolate multiple tenants in a cluster by default, overridable isolation
+
+Here is the existing tenancy related user story from our website:
+
+    As a cluster admin, I want to build tenants in my cluster that are isolated from each other by default. 
+    Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
+    may own more than 1 Namespace.
+
+The wording used here has led to much confusion. Specifically, the "by default" part actually means this policy should 
+be at the BANP priority, and solves the "Engineering org vs Marketing org" use case where you want to keep orgs 
+from accidentally interfering with each other. By default, cross-tenant traffic should be dropped. 
+However, namespace owners can override this behavior by applying their own policies as needed. 
+
+In reality, the CR attached to this story defines strict tenancy, which should be a separate use case (see next section).
+
+To make the use case more obvious we can add some details to it like the following
+
+    As a cluster admin, I want to build tenants for different departments (e.g. Engineering vs Marketing) 
+    in my cluster that are isolated from each other by default.
+    By default, cross-tenant traffic is dropped. However, namespace owners can override this behavior by applying 
+    their own policies as needed.
+    Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
+    may own more than 1 Namespace.
+
+#### Story 4.2: Create and Isolate multiple tenants in a cluster, strict isolation
+
+Strict tenancy is the "Coke vs Pepsi" sort of thing where you want each tenant to feel like it has its own cluster, 
+and be totally independent of the other tenants. We can write it down like this
+
+    As a cluster admin, I want to build tenants for different organizations (e.g. Coke vs Pepsi) 
+    in my cluster that are isolated from each other, where this isolation
+    can't be overridden by namespace owners. This policy should make every tenant completely independent and isolated 
+    from other tenants. Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
+    may own more than 1 Namespace.
+
+#### Story 4.3: Allow internal connections for tenants
+    
+    As a cluster admin, I want to build tenants in my cluster and always allow connections inside one tenant.
+    At the same time I want to setup an overridable deny-all policy to protect namespaces by default.
+    This policy should make sure internal connectivity for a tenant is always allowed, in case there are
+    lower-priority deny rules.
+
+#### Story 4.4: Tenants interaction with (B)ANP
+
+    As a cluster admin, I want to be able to setup policies with higher and lower priority than tenancy policy.
+    I want to deny inter-tenant connections, but I want to allow ingress traffic from the monitoring namespace
+    to all namespaces. Therefore, I need "allow from monitoring" rule to have higher priority than "deny from
+    other tenants".
+    I want to setup "deny all" BANP to protect cluster workloads, but I want to allow internal connections
+    within tenant. Therefore, I need "allow from same tenant" rule to have higher priority than "deny all" BANP.
+
+#### What I couldn't figure out user stories for
+
+- Skip action
+- Ports *[]AdminNetworkPolicyPort
+
+### Existing API
+
+AdminNetworkPolicy has the “SameLabels” and “NotSameLabels” fields to support the use cases involving tenancy. For example:
+
+**Use case**
+
+Traffic should be disallowed by default between namespaces owned by different users (defined by labels i.e `user=foo` or `user=bar`).
+
+**Policy**
+```
+kind: BaselineAdminNetworkPolicy
+apiVersion: policy.networking.k8s.io/v1alpha1
+metadata:
+  name: user-based-tenancy
+spec:
+  subject:
+    namespaces:
+      matchExpressions:
+        - key: user
+          operator: Exists
+  ingress:
+    - action: Deny
+      from:
+        - namespaces:
+            notSameLabels:
+              - user
+```
+
+**Meaning**
+
+In namespaces that have a “user” label, by default, deny ingress from namespaces that have a different “user” label.
+
+There are several major problems with this implementation of tenancy as it pertains to the user stories.
+
+First, There is no explicit definition of "tenancy" anywhere. The administrator has an idea of 
+"tenants are defined by the user label", but that's only true because this particular ANP happens to include that
+particular rule, and there's no way to find the ANP(s) that defines tenancy if you don't already know what they are.
+
+Second, the SameLabels/NotSameLabels selectors behave really differently from other peers, causing multiple underlying 
+rules to be created, and the syntax doesn't make that obvious.
+
+Third, the syntax is very general purpose / powerful. ANP has subjects and peers, which are different, 
+and currently Tenancy is defined on the peers side. Tenancy by itself has the same subject and peer, 
+at least for the existing use cases, and having separate selectors for subject and peer allows for more
+configurations than needed. 
+
+Fourth, the ANP subject allows using pod selectors, while tenancy use cases only need namespace selectors.
+
+## API
+
+TBD
+
+## Conformance Details
+
+TBD
+<!---
+(This section describes the names to be used for the feature or
+features in conformance tests and profiles.
+
+These should be `CamelCase` names that specify the feature as
+precisely as possible, and are particularly important for
+Extended features, since they may be surfaced to users.)
+-->
+
+## Alternatives
+
+Other alternatives were mentioned https://docs.google.com/document/d/113xBe7VMK7hMYdIdB9gobp7JwVkWQLnqdMPNkamfaK8/edit,
+but none of them cover all the Goals defined in this NPEP.
+
+There are 2 main problems with leaving Tenancy as a (B)ANP peer:
+1. tenancy is only based on namespace labels, but (B)ANP subject allows using pod selector too
+2. tenancy definition is less obvious, since it is a part of the peers list. Usually peer rules are the same for
+all (B)ANP subject pods, but for tenancy that is not true.
+3. There are actually more that 2 problems, mainly about allowing much more configurations for tenancy than we have
+user stories for, but it is covered in the previous sections :)
+
+Therefore, creating a new object seems like a more clear way to implement Tenancy.
+
+## References
+
+- https://docs.google.com/document/d/113xBe7VMK7hMYdIdB9gobp7JwVkWQLnqdMPNkamfaK8


npep/npep-122.md
```diff
@@ -0,0 +1,181 @@
...
+policy, it doesn’t really make it obvious that that’s what it’s doing.
+
+Second, the syntax is very general purpose / powerful. ANP has subjects and peers, which are different, 
+and currently Tenancy is defined on the peers side. Tenancy by itself has the same subject and peer, 
+at least for the existing use cases, and having separate selectors for subject and peer allows for more
+configurations than needed.
+
+## API
+
+We suggest to define a separate API for tenancy, as in the following example:
```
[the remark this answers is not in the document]
astoycos: (what you describe below are not new "high level" user stories

npep/npep-122.md
```diff
@@ -0,0 +1,181 @@
...
+
+```
+kind: AdminNetworkPolicyTenancy
+apiVersion: policy.networking.k8s.io/v1alpha1
+metadata:
+  name: user-based
+spec:
+  namespaceSelector:
```
astoycos, line 103: Maybe this would be `tenantSubject` or something? Although I don't really want to hash about API design just yet 🤣

npep/npep-122.md
```diff
@@ -0,0 +1,181 @@
...
+    matchExpression:
+      - key: system-namespace
+        operator: DoesNotExist
+      - key: user
+        operator: Exists
+  tenancyLabels:
```
astoycos, line 109: `tenancyDefinition` or just `definition` lol

astoycos (commented):
Still working through this but I've got the review mostly done.

npep/npep-122.md
```diff
@@ -0,0 +1,181 @@
...
+    > Example: say we define a tenancy based on `label1` and `label2`, and the following namespaces are selected by `spec.namespaceSelector`
+    > ```
+    > ns1{label1=A, label2=B, label3=C}
+    > ns2{label1=A, label2=B, label4=D}
+    > ns3{label1=_, label2=B}
+    > ns4{label1=_, label2=""}
+    > ns5{label1=_, label2=_}
+    > ns6{label1=A, label2=B}
+    > ```
+    > where `_` means the label is not set, and `""` means empty string value.
```
Dyanngg, line 135: We can touch upon naming later but I just what to point out that, this categorization of tenants is quite different than our original `sameLabels` proposal, where namespaces who do not have the tenancy label at all (as opposed to have an empty label) are not affected by the policy. 
When I read the `tenancyLabels` field, it is supposed to mean "namespaces that have the exact same values for these labels is considered a single tenant", but it does not and should not IMO imply that namespaces that are missing this label automatically form a tenant by themselves

npep/npep-122.md
```diff
@@ -0,0 +1,181 @@
...
+              - user
+```
+
+**Meaning**
+
+In namespaces that have a “user” label, by default, deny ingress from namespaces that have a different “user” label.
+
+There are two problems with this implementation of tenancy, relative to the user stories that it exists to support
+
+First, it never explicitly defines what a tenant is. Although the object correctly implements the desired tenancy
```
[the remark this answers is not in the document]
npinaeva: tried to rephrase that a bit

npep/npep-122.md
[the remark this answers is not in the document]
npinaeva: I actually thought we leave user stories (may need to decide if we need BANP-level tenancy though), but change how API implements them. That is also why this proposal has `Implementable` status, as I wasn't going to change user stories (but we can rework that of course, if you think that would be better)

npep/npep-122.md
npinaeva: Current design is supposed to be more explicit on which namespaces should be affected, and also allow different configurations. So if you don't want namespace without a "tenancy label" to be affected, you just add
`key: <tenancy label>, operator: Exists` to the `spec.namespaceSelector`. But at the same time if you do want to make a separate tenant for all namespaces without that label you leave empty selector.

Also I think `tenancyLabels` does exactly what you said: "namespaces that have the exact same values for these labels are considered a single tenant", because all namespaces with empty label will be one tenant, and all namespaces without a label (say they all have None value for it) will be one tenant. And you can easily avoid namespaces with unset labels as I pointed above.

danwinship: We could alternatively say that namespaces without all of the tenancyLabels are always excluded from the definition. IOW, `namespaceSelector` automatically implicitly includes `key: X, operator: Exists` for every label in `tenancyLabels`

npep/npep-122.md
```diff
@@ -0,0 +1,182 @@
...
+
+- third, define policy rules. This includes setting ports, action, and peer tenants. 
+For now, we only have `SameTenant` and `NotSameTenant` as a peer option.
+
+**Note** with the new tenancy definition there is no need to have `sameLabels` and `notSameLabels`, because
+these selectors will define the same tenants, but it is replaces with `SameTenant` and `NotSameTenant` for similar use cases.
+
+Existing API allows defining tenancy for BANP, we need to decide if the tenancy actually makes sense for that priority.
+BANP is overridable by the namespace owners, so the question is if there is a use case for a tenancy policy that can be
+overridden?
```
danwinship, line 154: There is. There are two kinds of "tenant" situation. One is the "Coke vs Pepsi" sort of thing where you want each tenant to feel like it has its own cluster, and be totally independent from the other tenants. In that case you want hard/non-overridable tenancy. But the other case is the "Engineering vs Marketing" or "OpenShift team vs RHEL team" sort of tenancy, where you want to keep people from _accidentally_ interfering with each other, but you want to allow them to communicate with each other if they want to. So for that you want isolated-by-default, but overridable.

(I think as multi-cluster becomes better supported, and things like HyperShift allow you to share resources between multiple clusters, there will be fewer people trying to do hard tenancy within a single cluster.)

npep/npep-122.md
npinaeva: We can do that, but I see 2 possible disadvantages to this:
- `namespaceSelector` doesn't fully define which namespaces will be selected, so we add an implicit label selector `key: X, operator: Exists` and it makes the API more confusing
- there is no way to set rules for namespaces with label unset: e.g. if I want to deny-from-not-same-tenant, I may want to deny from not only namespaces with different label value, but also from namespaces without a label, otherwise forgetting to set a label on a namespace may become a security hole, because it won't be affected by tenancy.

npep/npep-122.md
```diff
@@ -0,0 +1,182 @@
...
+  ingress:
+    - action: Deny
+      from:
+        - NotSameTenant
+```
+
+that will define a tenant as follows:
+- first, select namespaces that will be split into tenants, and have the defined policy applied. `spec.namespaceSelector`
+is used for this purpose. This also solves the problem we had for empty or unset labels: if tenancy is defined based on label `user`,
+and namespaces without this label should not be affected, `key: user, operator: Exists` selector can be used.
```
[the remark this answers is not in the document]
npinaeva: yeah those 2 steps work together, but here I meant that there is not implicit selector for namespaces (like all labels from tenancyLabels should exist), so all namespace selected here will be affected, and how they are split into tenants in the next step

npep/npep-122.md
npinaeva: I find these use cases very easy to understand, so I added user stories clarifications section

npep/npep-122.md
```diff
@@ -0,0 +1,215 @@
...
+at least for the existing use cases, and having separate selectors for subject and peer allows for more
+configurations than needed. Another point to this is the ANP subject allows using pod selectors, while tenancy
+use cases only need namespace selectors.
+
+## API
+
+We suggest to define a separate API Object for tenancy, as in the following example:
+
+```
+kind: AdminNetworkPolicyTenancy
```
astoycos, line 136:
```suggestion
kind: AdminNetworkPolicyTenant
```

npep/npep-122.md
```diff
@@ -0,0 +1,215 @@
...
+"tenants are defined by the user label", but that's only true because this particular ANP happens to include that
+particular rule, and there's no way to find the ANP(s) that defines tenancy if you don't already know what they are.
+
+Second, SameLabels/NotSameLabels behaves really differently from other peers, causing multiple underlying 
+rules to be created, and the syntax doesn't make that obvious.
+
+Third, the syntax is very general purpose / powerful. ANP has subjects and peers, which are different, 
+and currently Tenancy is defined on the peers side. Tenancy by itself has the same subject and peer, 
+at least for the existing use cases, and having separate selectors for subject and peer allows for more
+configurations than needed. Another point to this is the ANP subject allows using pod selectors, while tenancy
```
astoycos, line 128: Make this "other point" a Fourth... I think that the current def allows you to select pods even though that wasn't specified in our user stories is a major issue.

astoycos (commented):
Looking good! I really like the way you explicitly laid out the issues with the existing implementation. I'm planning on lgtm(ing) this KEP without focusing too hard on the actual "API" section as I Think that should be completed in another PR.  Let's try to build consensus around the updated user stories and various issues you pointed out in the existing implementation first.

npinaeva:
As discussed in the upstream meeting, this version is focused on user stories now, which led me to writing down even more user stories, and also covering existing API fields that I couldn't figure out user stories for.
Since the main problem of the existing API is covering much more configs that we actually have user stories for, we probably shouldn't include fields that we don't have user stories for in the next Tenancy API.

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+
+**Meaning**
+
+In namespaces that have a “user” label, by default, deny ingress from namespaces that have a different “user” label.
+
+There are two major problems with this implementation of tenancy as it pertains to the user stories.
+
+First, There is no explicit definition of "tenancy" anywhere. The administrator has an idea of 
+"tenants are defined by the user label", but that's only true because this particular ANP happens to include that
+particular rule, and there's no way to find the ANP(s) that defines tenancy if you don't already know what they are.
```
rahulkjoshi, line 150: While this is true, I'm not sure this is a huge problem unique to tenancy. I could make a parallel argument for something like egress rules: There's no way to find the ANP(s) that define egress if you don't already know what they are.

I would argue that this is a problem easily solved by a good naming scheme (e.g. call my ANP "tenant-policy" or "namespace-isolation"). Easy discoverability of a particular type of rule shouldn't be a top-priority for our design, in my opinion.

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+        - namespaces:
+            notSameLabels:
+              - user
+```
+
+**Meaning**
+
+In namespaces that have a “user” label, by default, deny ingress from namespaces that have a different “user” label.
+
+There are two major problems with this implementation of tenancy as it pertains to the user stories.
```
rahulkjoshi, line 146:
```suggestion
There are several major problems with this implementation of tenancy as it pertains to the user stories.
```

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+
+Second, the SameLabels/NotSameLabels selectors behave really differently from other peers, causing multiple underlying 
+rules to be created, and the syntax doesn't make that obvious.
+
+Third, the syntax is very general purpose / powerful. ANP has subjects and peers, which are different, 
+and currently Tenancy is defined on the peers side. Tenancy by itself has the same subject and peer, 
+at least for the existing use cases, and having separate selectors for subject and peer allows for more
+configurations than needed. 
+
+Fourth, the ANP subject allows using pod selectors, while tenancy use cases only need namespace selectors.
```
rahulkjoshi, line 160: Is this a hard requirement? If a admin wants to isolate Pods by a label is that a valid use-case? Maybe we can add it, but I don't think thats too unreasonable and I don't think it is unimplementable (citation needed).

For example: I want to define my tenant as the value of the label `app` and deny Pods with that label from trying to talk to anything outside their tenant.

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+In namespaces that have a “user” label, by default, deny ingress from namespaces that have a different “user” label.
+
+There are two major problems with this implementation of tenancy as it pertains to the user stories.
+
+First, There is no explicit definition of "tenancy" anywhere. The administrator has an idea of 
+"tenants are defined by the user label", but that's only true because this particular ANP happens to include that
+particular rule, and there's no way to find the ANP(s) that defines tenancy if you don't already know what they are.
+
+Second, the SameLabels/NotSameLabels selectors behave really differently from other peers, causing multiple underlying 
+rules to be created, and the syntax doesn't make that obvious.
```
rahulkjoshi, line 153: I agree, I'm just curious from your perspective do you think this is bad from:
- a scale perspective, where the CNI might get over-whlemed by the number of rules
- unintended behavior perspective, where the user doesn't understand what they've done
- both

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+    Higher priority to allow/deny some inter-tenant connections, for example, allow ingress traffic from the
+    monitoring namespace without including that namespace to every tenant. Lower priority to allow/deny
+    connections that are not a part of tenant policy, for example, allow internal tenant connections with tenancy
+    and deny-all policy with lower priority.
+
+#### Story 4.5: Ensure every namespace in the cluster is a part of some tenant
+
+    As a cluster admin, I want to implement security zones tenants, and deny connections between different zones.
+    I want to use a label e.g. `security-zone` to build tenants, and I want to make sure every namespace in the cluster
+    is a part of some tenant/security zone. This way, I can be sure that even if the `security-zone`
```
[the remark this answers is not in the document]
npinaeva: Yes it is similar to 4.2, but this one stresses the "none" tenant as a requirement. Currently, we don't have a way to create "none" tenant, that is why I added an extra story to explain why this functionality may be needed explicitly

npep/npep-122.md
npinaeva: Very valid point, I don't think this is the major problem, but more of an introduction to the following points, just stressing that tenants are different from the other peer types. As you said, if you want to define a tenant, you will probably create a separate ANP with a single tenancy rule and a tenancy-specific name, but that is not dictated by the API, and tenancy may easily be lost in a list of other peers.

npep/npep-122.md
npinaeva: mostly `unintended behavior perspective` part

npep/npep-122.md
npinaeva: It should be implementable indeed, but that adds extra complexity for the API and we don't have any use cases for it yet. Another concern is that pod labels can usually be changed by the namespace owner, in that case admin can't enforce proper tenancy isolation base don pod labels.

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+
+Strict tenancy is the "Coke vs Pepsi" sort of thing where you want each tenant to feel like it has its own cluster, 
+and be totally independent of the other tenants. We can write it down like this
+
+    As a cluster admin, I want to build tenants in my cluster that are isolated from each other and this can't
+    be overridden. This policy should make every tenant completely independent and isolated from another tenants.
+    Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
+    may own more than 1 Namespace.
+
+#### Story 4.3: Allow internal connections for tenants
```
astoycos, line 85: Hrm this one seems a bit duplicative to the "create and isolate" stories to me...

More specifically... 

> At the same time I want to setup an overridable deny-all policy to protect namespaces by default.
    This policy should make sure internal connectivity for a tenant is always allowed, in case there are
    lower-priority deny rules.

In this case (thinking in terms on ANP) you'd use an ANP to setup the tenants... and that ANP would implicity be a "higher-order" policy meaning even with an "overridable deny-all policy" traffic within tenants would always be allowed unless explicitly denied with another ANP or "passed" to namespace owners

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+
+    As a cluster admin, I want to be able to setup policies with higher and lower priority than tenancy policy.
+    Higher priority to allow/deny some inter-tenant connections, for example, allow ingress traffic from the
+    monitoring namespace without including that namespace to every tenant. Lower priority to allow/deny
+    connections that are not a part of tenant policy, for example, allow internal tenant connections with tenancy
+    and deny-all policy with lower priority.
+
+#### Story 4.5: Ensure every namespace in the cluster is a part of some tenant
+
+    As a cluster admin, I want to implement security zones tenants, and deny connections between different zones.
```
astoycos, line 102: What's a "security zones tenants"?  Isn't a "zone" the same thing as a "tenant" basically?

npep/npep-122.md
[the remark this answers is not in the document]
astoycos: Hrm this seems dangerous to me... almost like the Implicit deny provided by network policy...

i.e we're saying here that when there's no tenants defined nothing happens... but the moment there's at least on tenant ALL non selected workloads are/ can be implicitly selected like a tenant now? 

i.e I think we need to stay away from use cases where rules are applied to workloads even if they aren't explicitly selected by a policy  

maybe I'm overthinking it though 🤷

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+    I want to use a label e.g. `security-zone` to build tenants, and I want to make sure every namespace in the cluster
+    is a part of some tenant/security zone. This way, I can be sure that even if the `security-zone` 
+    label is not assigned to the tenant for some reason, tenancy rules will still apply and security zones won't be
+    violated.
+
+#### What I couldn't figure our user stories for
+
+- Skip action
```
astoycos, line 110: hrm I feel like a skip action could help with 4.3 and 4.4 

"overridable deny-all policy to protect namespaces by default."

Could be implemented with a ANP + pass rules OR a BANP 

IMO at the end of the day it's hard because we're thinking about implementation details when writing user stories 🤣  even though a story can be a bit open ended leaving the design up to the implementer

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+    AdminNetworkPolicy Pass rules allows an admin to delegate security posture for certain traffic to the Namespace 
+    owners by overriding any lower precedence Allow or Deny rules. For example, intra-tenant traffic management can be 
+    delegated to tenant admins explicitly with the use of Pass rules.
+    
+So really, the only solidly-agreed-upon use case is that you should be able to create enforced isolation 
+between particular sets of namespaces.
+
+### Clarifying user stories
+
+#### Story 4.1: Create and Isolate multiple tenants in a cluster by default
```
danwinship, line 53: maybe add "(overridable isolation)" to the end of the heading, to parallel the "strict isolation" in the next heading

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+    Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
+    may own more than 1 Namespace.
+
+#### Story 4.2: Create and Isolate multiple tenants in a cluster, strict isolation
+
+Strict tenancy is the "Coke vs Pepsi" sort of thing where you want each tenant to feel like it has its own cluster, 
+and be totally independent of the other tenants. We can write it down like this
+
+    As a cluster admin, I want to build tenants in my cluster that are isolated from each other and this can't
+    be overridden. This policy should make every tenant completely independent and isolated from another tenants.
```
danwinship, line 81:
```suggestion
    As a cluster admin, I want to build tenants in my cluster that are isolated from each other where this
    isolation can't be overridden by the tenants themselves. This policy should make every tenant completely independent and isolated from another tenants.
```

(because the admin could still override the isolation with ANPs)

alternatively "where the isolation can't be overridden by NetworkPolicy". Actually, that might be a good way to describe both the previous case and this one; the previous case is where the tenant isolation _can_ be overridden by NetworkPolicy and this case is where it _can't_ be.

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+    Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
+    may own more than 1 Namespace.
+
+#### Story 4.3: Allow internal connections for tenants
+    
+    As a cluster admin, I want to build tenants in my cluster and always allow connections inside one tenant.
+    At the same time I want to setup an overridable deny-all policy to protect namespaces by default.
+    This policy should make sure internal connectivity for a tenant is always allowed, in case there are
+    lower-priority deny rules.
```
danwinship, line 90: Agree with Andrew; I'm not exactly sure what this one is saying. It sounds like you're saying the admin wants to forbid users from using NP within a tenant

npep/npep-122.md
danwinship: > "overridable deny-all policy to protect namespaces by default."
> 
> Could be implemented with a ANP + pass rules OR a BANP

How would you do overridable deny-all with Pass? If the rule Denies then it's not overridable and if it Passes then it's not denying

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+    As a cluster admin, I want to implement security zones tenants, and deny connections between different zones.
+    I want to use a label e.g. `security-zone` to build tenants, and I want to make sure every namespace in the cluster
+    is a part of some tenant/security zone. This way, I can be sure that even if the `security-zone` 
+    label is not assigned to the tenant for some reason, tenancy rules will still apply and security zones won't be
+    violated.
+
+#### What I couldn't figure our user stories for
+
+- Skip action
+- Ports *[]AdminNetworkPolicyPort
```
danwinship, line 111: Yeah, I don't think there's a good argument for using ports in tenant definitions.

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+- Multiple tenant policies (e.g. based on different labels?)
```
danwinship, line 112: I think we want to make this a Non-Goal. The existing KEP allows for having lots of rules with different SameLabels/NotSameLabels doing different things, but the user stories all involve having a single cluster-wide definition of "tenancy".

npep/npep-122.md
danwinship: > If a admin wants to isolate Pods by a label is that a valid use-case?

But this isn't about merely isolating Pods, it's about defining _tenancy_ at the Pod level. It's one thing to have an ANP that says "the frob operator should be able to access any pod with the 'frobbable' label in any Namespace". But it would be weird to say "all pods with the 'pod-tenant' label can talk to all other pods with the same valued 'pod-tenant' label in all namespaces", particularly since, as Nadia said, the users can just relabel their pods themselves whenever they want.

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
+# NPEP-122: NPEP template
```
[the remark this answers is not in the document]
shaneutt: Perhaps:

```suggestion
# NPEP-122: Tenancy
```

?

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Goals
+
+- Clarify tenancy use cases.
+- Provide a complete definition of a tenant in respect to ANP.
+- Separate ANP subject from Tenancy subject/peer.
+
+## Non-Goals
+
+- cover new use cases for Tenancy.
```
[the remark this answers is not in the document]
danwinship: I think the thought is that we had a pretty clear idea of what tenancy was supposed to be about, but we didn't explain it very well in the KEP. So explaining our existing idea better is a Goal, but coming up with new ideas is a Non-Goal.

npep/npep-122.md
npinaeva: It is very similar indeed, but the difference is in the action: "Isolate" user story is focused on the Deny tenancy rules, and "Allow internal connections for tenants" is more about  Allow tenancy rules. Maybe they are too similar, lmk if you think I should remove this one

npep/npep-122.md
npinaeva: yeah "security zone" was supposed to explain how the tenants are used, adding more semantics to the plain tenants for this user story. Tried to reword it a bit to make it more obvious

npep/npep-122.md
npinaeva: I agree we shouldn't focus too much on this, but since I did spend some time trying to figure out a use case for it and couldn't, I decided to write it down :)
Skip action is mostly used to optimize interaction between different ANPs (if we consider tenancy and ANP to be different things in this context), so that it allows to ignore the ANPs with lower-level priorities (ANP<->ANP relations). We know that every skip action can be replaced with just not selecting Skipped pods in the netpols with lower prio, but skipping allows more flexibility. From that perspective, I don't think it makes sense for tenancy to define a skip action for other ANPs (tenancy<->ANP), same as I don't think there is a reasonable use case to skip tenancy with the lower priority (tenancy<->tenancy).
lmk if that makes any sense :)

npep/npep-122.md
```diff
@@ -0,0 +1,194 @@
+# NPEP-122: NPEP template
+
+* Issue: [#122](https://github.com/kubernetes-sigs/network-policy-api/issues/122)
+* Status: Provisional
+
+## TLDR
+
+Stop using SameLabels/NotSameLabels to define tenancy within ANP. Instead, propose a separate object
+that will define Tenancy and ensure that subject and peer relation is explicitly defined.
```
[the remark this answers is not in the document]
npinaeva: I agree too :) there may be some artifacts like that from the first version of npep that was focused on the specific API changes instead of user stories, let me clean this up

npep/npep-122.md
npinaeva: good point! I made it "namespace owners" just based on the first user story wording, lmk if you prefer some other name for that actor :)

npep/npep-122.md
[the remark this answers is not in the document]
Dyanngg: I completely agree with your statements above. I think the confusion though comes from whether this is a "use case" v.s. "api semantics clarification". IMO, we should actually provide sample tenant-rule policy which only selects namespaces that contains the tenancy label(s) as a best practice. In addition, we clarify that if "extra" namespaces are selected, what's the expected behavior, and _maybe_ that is not recommended. But the other way around does not make sense to me: because I want to have every namespace in the cluster strictly associated with a tenant, let's select all namespaces in the policy while I could have just selected the "tenant namespaces" that I care about.

npep/npep-122.md
Dyanngg: Please take a look at the second example yaml in https://network-policy-api.sigs.k8s.io/reference/examples/#sample-spec-for-story-4-create-and-isolate-multiple-tenants-in-a-cluster. This is how we originally envisioned `Pass` is used in creating tenancy rules

npinaeva: very nice link, thank you @Dyanngg! 
I am not sure I understand that example: why would you write 
```
apiVersion: policy.networking.k8s.io/v1alpha1
kind: AdminNetworkPolicy
metadata:
  name: tenant-creation-example
spec:
  priority: 50
  subject:
    namespaces:
      matchExpressions: {key: "tenant"; operator: Exists}
  ingress:
    - action: Pass # Pass inter-tenant traffic to any defined NetworkPolicies
      from:
      - namespaces:
          sameLabels:
          - tenant
    - action: Deny   # Deny everything else other than same tenant traffic
      from:
      - namespaces:
          namespaceSelector: {}
```
when it is equivalent to
```
apiVersion: policy.networking.k8s.io/v1alpha1
kind: AdminNetworkPolicy
metadata:
  name: tenant-creation-example
spec:
  priority: 50
  subject:
    namespaces:
      matchExpressions: {key: "tenant"; operator: Exists}
  ingress:
    - action: Deny   # Deny everything else other than same tenant traffic
      from:
      - namespaces:
          namespaceSelector: {}
```
and doesn't seem to actually do what the user story wanted? Maybe I am missing something here?

rahulkjoshi: I think I see @Dyanngg's point -- by putting in an explicit `Pass` directive, you make the rule more precise.

quick nit: I think if we want to drop the pass, the rule shuold be:
```
apiVersion: policy.networking.k8s.io/v1alpha1
kind: AdminNetworkPolicy
metadata:
  name: tenant-creation-example
spec:
  priority: 50
  subject:
    namespaces:
      matchExpressions: {key: "tenant"; operator: Exists}
  ingress:
    - action: Deny   # Deny everything else other than same tenant traffic
      from:
      - namespaces:
          notSameLabels:
          - tenant
```

While the example you provide does indeed achieve the same in isolation, what happens if you have multiple rules. For example a rule at priority 100 (evaluated after the rule at priority 50) could allow traffic between pods with the label `my-app: example`. By having the pass rule at priority 50, they're able to say "I've defined a tenancy relationship at rule 50. Inter-tenant traffic is dropped, and traffic within a tenant is up to the namespace admins. Now for anything not selected by that tenancy rule, allow traffic between apps with the label `my-app: example`"

While the toy example of `my-app: example` is a little contrived, a cluster admin shouldn't have to always put in an exception if they want to provide a rule with lower precedence than a tenancy rule (i.e. rule at 100 needs to explicitly not select the tenant traffic).

npinaeva: @rahulkjoshi your yaml is absolutely correct, and it is actually listed as the first option https://network-policy-api.sigs.k8s.io/reference/examples/#sample-spec-for-story-4-create-and-isolate-multiple-tenants-in-a-cluster
so the question I have is why we need the second option, because I am trying to figure out if Pass action makes sense for tenancy in general, and in this second option yaml in particular

npep/npep-122.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Story 4.1: Create and Isolate multiple tenants in a cluster by default, overridable isolation
+
+Here is the existing tenancy related user story from our website:
+
+    As a cluster admin, I want to build tenants in my cluster that are isolated from each other by default. 
+    Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
+    may own more than 1 Namespace.
+
+The wording used here has lead to much confusion. Specifically, "By default" part actually means this policy should
```
astoycos, line 63:
```suggestion
The wording used here has lead to much confusion. Specifically, the "by default" part actually means this policy should 
```

npep/npep-122.md
```diff
@@ -0,0 +1,188 @@
+# NPEP-122: Tenancy API
+
+* Issue: [#122](https://github.com/kubernetes-sigs/network-policy-api/issues/122)
+* Status: Provisional
+
+## TLDR
+
+Tenancy definition and the API (SameLabels/NotSameLabels) is confusing and ambiguous as of now. 
+We want to rethink tenancy use cases and the API to keep it simple and easy to understand, but
```
astoycos, line 9:
```suggestion
We want to rethink the tenancy use cases and the API to keep it simple and easy to understand, but
```

npep/npep-122.md
```diff
@@ -0,0 +1,215 @@
...
+
+"By default" part actually means this policy should be at the BANP priority, and solves "Engineering vs Marketing"
+use case where you want to keep people from accidentally interfering with each other, 
+but you want to allow them to communicate with each other. But the CR attached to this story actually defines
+strict tenancy, which should be a separate use case (see next section).
+
+To make the use case more obvious we can add some details to it like the following
+
+    As a cluster admin, I want to build tenants in my cluster that are isolated from each other by default.
+    This should prevent tenants from interfering with each other by accident, but may be overridden by namespace
```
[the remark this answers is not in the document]
astoycos: @npinaeva Can you make this ^ change?

npep/npep-122.md
astoycos: > "Isolate" user story is focused on the Deny tenancy rules, and "Allow internal connections for tenants" is more about Allow tenancy rules.

IMO this is somewhat implicit in:

```
As a cluster admin, I want to build tenants in my cluster that are isolated from each other where this isolation
    can't be overridden by namespace owners.
```

i.e we're building hard tenants where the definition is tenants can't talk to each other but members of a tenant are of one group, and can communicate.

If you want to augment 4.2 to explicitly mention something like that I think that'd be alright, but I think Dan and I agree here so I'd remove 4.3

```
   As a cluster admin, I want to build tenants in my cluster that are isolated from each other where this isolation
    can't be overridden by namespace owners. This policy should make every tenant completely independent and isolated 
    from another tenants, while explicitly allowing same tenant members to communicate.
 ```

File: npep/npep-122.md (hunk 17 above)
Comment (astoycos) [reply npep/npep-122.md:105]: I think this is a good section leave... however on the API side of things I think it's alright to still possibly use pass in whatever we come up with during the design stage.... remember we've agreed that it's OK not to have an explicit user story for re-applying a feature that's already in the api, However, if we chose to DISALLOW pass rules when building tenants then we will need some explicit documentation that describes why.

**Review by astoycos — APPROVED** [review]
/lgtm
/hold

File: npep/npep-122.md (hunk 15 above)
Comment (npinaeva) [reply npep/npep-122.md:88]: let me try to explain this once again, because I don't think we are talking about the same use case

4.2 story just denies inter-tenant communication, but doesn't say anything about intra-tenant communication, e.g. you could write "deny from the other tenants" ANP, and at the same time define a NP in namespace A saying "deny from namespace B", where namespaces A and B are in the same tenant. That would limit intra-tenant communication, and ANP doesn't say anything about that, so traffic from namespace B will be dropped.

4.3 use case is aimed to prevent namespaces in the same tenant from denying each others connections (e.g. by applying any network policy and by mistake not including your tenant namespaces to the allow list). In that case even if same namespace A applies a "deny from ALL" network policy, namespace B will still be allowed by ANP.

I just want to make sure I've explained the use cases well enough, and then we can surely remove any extra parts or whole user stories.

File: npep/npep-122.md (hunk 17 above)
Comment (npinaeva) [reply npep/npep-122.md:105]: exactly, this section is here just to be explicit on the use case, and doesn't imply anything for the API itself

File: npep/npep-122.md [hunk 29]
```diff
@@ -0,0 +1,188 @@
...
+
+#### Story 4.1: Create and Isolate multiple tenants in a cluster by default, overridable isolation
+
+Here is the existing tenancy related user story from our website:
+
+    As a cluster admin, I want to build tenants in my cluster that are isolated from each other by default. 
+    Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
+    may own more than 1 Namespace.
+
+The wording used here has lead to much confusion. Specifically, the "by default" part actually means this policy should
```
danwinship, line 63:
```suggestion
The wording used here has led to much confusion. Specifically, the "by default" part actually means this policy should 
```

npep/npep-122.md
```diff
@@ -0,0 +1,188 @@
...
+    may own more than 1 Namespace.
+
+#### Story 4.2: Create and Isolate multiple tenants in a cluster, strict isolation
+
+Strict tenancy is the "Coke vs Pepsi" sort of thing where you want each tenant to feel like it has its own cluster, 
+and be totally independent of the other tenants. We can write it down like this
+
+    As a cluster admin, I want to build tenants in my cluster that are isolated from each other where this isolation
+    can't be overridden by namespace owners. This policy should make every tenant completely independent and isolated 
+    from another tenants. Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant
```
danwinship, line 85:
```suggestion
    from other tenants. Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
```

npep/npep-122.md
danwinship: > 4.3 use case is aimed to prevent namespaces in the same tenant from denying each others connections

Is that a real use case? Why would an administrator want to enforce that for every tenant at the cluster level?

4.1 is "Engineering vs Marketing" and 4.2 is "Coke vs Pepsi". What is the explanation for this user story?

(Also, you should probably actually put "Engineering vs Marketing" and "Coke vs Pepsi" into the actual user stories. The user story should stand alone to explain the use case. We shouldn't have the user story _and_ also have a separate "here's what this really means". And in this case (4.3), I don't think there _is_ any "here's what this really means".)

npep/npep-122.md
```diff
@@ -0,0 +1,188 @@
...
+#### Story 4.3: Allow internal connections for tenants
+    
+    As a cluster admin, I want to build tenants in my cluster and always allow connections inside one tenant.
+    At the same time I want to setup an overridable deny-all policy to protect namespaces by default.
+    This policy should make sure internal connectivity for a tenant is always allowed, in case there are
+    lower-priority deny rules.
+
+#### Story 4.4: Tenants interaction with (B)ANP
+
+    As a cluster admin, I want to be able to setup policies with higher and lower priority than tenancy policy.
```
danwinship, line 97:
```suggestion
    As a cluster admin, I want to be able to setup policies with higher and lower priority than tenant policy.
```

I think? You're saying you want policies both higher and lower than the policies that the tenants define for themselves, not that you want policies both higher and lower than "the policy that defines what tenants are" or something.

npep/npep-122.md
```diff
@@ -0,0 +1,188 @@
...
+    Higher priority to allow/deny some inter-tenant connections, for example, allow ingress traffic from the
+    monitoring namespace without including that namespace to every tenant. Lower priority to allow/deny
+    connections that are not a part of tenant policy, for example, allow internal tenant connections with tenancy
+    and deny-all policy with lower priority.
```
danwinship, line 101: The wording here is vague / too terse. I don't understand what you're saying.

Does "without including that namespace to every tenant" mean "without having to write a separate policy for every tenant"? If so, I'd suggest that wording.

I can't figure out what the lower priority example means at all.

npep/npep-122.md
```diff
@@ -0,0 +1,188 @@
...
+Third, the syntax is very general purpose / powerful. ANP has subjects and peers, which are different, 
+and currently Tenancy is defined on the peers side. Tenancy by itself has the same subject and peer, 
+at least for the existing use cases, and having separate selectors for subject and peer allows for more
+configurations than needed. 
+
+Fourth, the ANP subject allows using pod selectors, while tenancy use cases only need namespace selectors.
+
+## API
+
+TBD
```
danwinship, line 158: (Not for this PR, but: I'm thinking tenancy should just be defined by a single Namespace label. You can use `kubernetes.io/metadata.name` if you want each namespace to be its own tenant, or `user` or `company` or `department` or whatever if you want to define tenancy with groups of namespaces.)

npep/npep-122.md
npinaeva: I tried to re-word it, ptal

npep/npep-122.md
npinaeva: I thought about adding multiple-label-based tenancy to `What I couldn't figure out user stories for` section, but wasn't sure if want to change that

astoycos:
/lgtm

🍾

Issue #127, opened 2023-07-19
[ENHANCEMENT] Add Southbound traffic support for ANP API


npinaeva:
**Is your enhancement request related to a problem? Please describe.**
This issue should track ingress policy use cases from external endpoints. nodes and hostNetwork pods. This is another piece to the similar egress-only discussion https://github.com/kubernetes-sigs/network-policy-api/issues/126

**Describe the solution you'd like**
Add (B)ANP fields to specify ingress peers that are different from cluster-network pods.

**Describe alternatives you've considered**
Keep using NetworkPolicy for admin needs :)

LiorLieberman:
Has this feature ever been implemented? or are there any current plans to implement that? saw nadia's pr from september.

PR #128, opened 2023-07-20
Create NPEP-127 for external ingress endpoints

npinaeva:
An attempt to collect use cases for external ingress endpoints.
Reviews and more examples are welcome.

Tracking issue https://github.com/kubernetes-sigs/network-policy-api/issues/127

### DIFF
--- a/npep/npep-127.md
+++ b/npep/npep-127.md
@@ -1,0 +1,90 @@
+# NPEP-127: Add southbound traffic support in (B)ANP API
+
+* Issue: [#127](https://github.com/kubernetes-sigs/network-policy-api/issues/127)
+* Status: Provisional
+
+## TLDR
+
+This NPEP proposes adding support for cluster ingress (southbound) traffic control in the `AdminNetworkPolicy` and `BaselineAdminNetworkPolicy` API objects.
+
+## Goals
+
+* Implement ingress traffic control from external destinations (outside the cluster)
+
+## Non-Goals
+
+(What is out of scope for this proposal.)
+
+## Introduction
+
+Currently, (B)ANP only allows to specify [peers](https://github.com/kubernetes-sigs/network-policy-api/blob/46be90096f23934f56342bc408b94a1aa79a159c/apis/v1alpha1/shared_types.go#L98-L103) 
+based on pod and namespace selector, that will only apply to non-hostNetwork cluster pods. This enhancement aims to
+add more types of incoming connections that may be specified. A similar effort exists for the [egress](https://github.com/kubernetes-sigs/network-policy-api/issues/126)
+direction.
+
+By default, cluster workloads are not accessible from outside the cluster, but there are many ways to expose these workloads
+(either with k8s APIs like [Ingress API](https://kubernetes.io/docs/concepts/services-networking/ingress/) or
+[Gateway API](https://gateway-api.sigs.k8s.io/), or with downstream implementations,
+like [Openshift's Route CR](https://docs.openshift.com/container-platform/4.13/rest_api/network_apis/route-route-openshift-io-v1.html)).
+Some of the APIs may provide functionality to filter incoming connections, but Network Policy API is different, because
+it is applied to the cluster workloads, and should provide the required level of security regardless of how/if a cluster workload
+is exposed to the outer world.
+
+### Use cases
+Most of the use cases refer to ANP priority, meaning the policy should be non-overridable by namespace owners, unless
+other clarification are mentioned in the use case itself.
+
+ - zero trust policy
+
+    As a cluster administrator when implementing a zero trust policy I want to make sure no ingress connection 
+    (unless explicitly allowed) will get to the cluster workloads. This should apply to all connecitons, including traffic from outside the cluster.
+
+ - block well-known ports
+
+    As a cluster administrator I decided that some well-known services like ftp, telnet, SNMP, etc. 
+    should not be allowed for cluster workloads. To implement this policy I want to explicitly deny all ingress connections
+    for TCP ports 21, 23, 161, etc.
+
+ - compromised node protection
+ 
+    My cluster has a set of nodes with very sensitive workloads, as a cluster administrator I want to make sure that if 
+    a worker node A is compromised, it will not be able to affect other worker nodes. To do so, I want to create a cluster-wide policy to
+    deny ingress connections from some of the cluster nodes. I may also need to explicitly allow access from the control plane nodes.
+    This security model assumes all pods from the same namespace are scheduled to the same node, and ingress connections
+    from pods located on the compromised node may be denied with existing namespace selectors.
+
+ - allow external endpoints
+
+    I have a cluster (management cluster) that creates other clusters on request (hosted clusters).
+    Management cluster assigns a subnet for hosted cluster workloads. As a cluster administrator I want to allow
+    incoming connections from the hosted clusters identifying them with assigned CIDRs
+
+
+- external services
+
+    I have an external service to collect telemetry and metrics from multiple clusters.
+    As a cluster administrator I want to make sure ingress connections from this service can't be denied by the namespace owners 
+    with NetworkPolicy.
+
+## API
+
+(... details, can point to PR with changes)
+
+## Conformance Details
+
+(This section describes the names to be used for the feature or
+features in conformance tests and profiles.
+
+These should be `CamelCase` names that specify the feature as
+precisely as possible, and are particularly important for
+Extended features, since they may be surfaced to users.)
+
+## Alternatives
+
+(List other design alternatives and why we did not go in that
+direction)
+
+## References
+
+(Add any additional document links. Again, we should try to avoid
+too much content not in version control to avoid broken links)


npep/npep-127.md
```diff
@@ -0,0 +1,88 @@
...
+
+Currently, (B)ANP only allows to specify [peers](https://github.com/kubernetes-sigs/network-policy-api/blob/46be90096f23934f56342bc408b94a1aa79a159c/apis/v1alpha1/shared_types.go#L98-L103) 
+based on pod and namespace selector, that will only apply to non-hostNetwork cluster pods. This enhancement aims to
+add more types of incoming connections that may be specified. A similar effort exists for the [egress](https://github.com/kubernetes-sigs/network-policy-api/issues/126)
+direction.
+
+By default, cluster workloads are not accessible from outside the cluster, but there are many ways to expose these workloads
+(either with k8s APIs like [Ingress API](https://kubernetes.io/docs/concepts/services-networking/ingress/) or
+[Gateway API](https://gateway-api.sigs.k8s.io/), or with downstream implementations,
+like [openshift Route](https://docs.openshift.com/container-platform/4.13/rest_api/network_apis/route-route-openshift-io-v1.html)).
```
astoycos, line 28:
```suggestion
like [Openshift's Route CR](https://docs.openshift.com/container-platform/4.13/rest_api/network_apis/route-route-openshift-io-v1.html)).
```

npep/npep-127.md
```diff
@@ -0,0 +1,88 @@
...
+is exposed to the outer world.
+
+### Use cases
+ - zero trust policy: As a cluster administrator I want to set up default deny all ingress policy as a BANP.
+
+    Example: when implementing a zero trust policy I want to make sure no ingress connection (unless explicitly allowed) will
+get to the cluster workloads. This should apply not only to the connections coming from the other cluster workloads, but to
+all incoming connections.
+
+ - block well-known ports: As a cluster administrator I want to block all ingress traffic on specific ports.
```
astoycos, line 40: Will/should this apply to well-known ports on Pods AND nodes?

npep/npep-127.md
```diff
@@ -0,0 +1,88 @@
...
+
+ - inter-cluster communication: As a cluster manager for multiple clusters I want to make sure ingress connections
+are only allowed from a pre-defined set of related clusters, which are identified by their network CIDR.
+
+    Example: I have a cluster (management cluster) that creates other clusters on request (hosted clusters).
+Management cluster assigns a subnet for hosted cluster workloads. I want to make sure that hosted cluster workloads
+have limited access to the management cluster workloads by selecting management cluster workloads I want to protect
+and a set of allowed subnets for the corresponding hosted clusters.
+
+- external services: As a cluster manager I want to make sure ingress connections from a pre-defined set of external endpoints
```
astoycos, line 61: This is broad "pre-defined set of external endpoints" what's defined as an external endpoint? can it be ANY IP?

Also this is the first time I see "can't be denied by the namespace owners" can you add a note somewhere that all the other user stories are explicitly non-overridable by namespace owners?

astoycos (commented):
I think we still have some thinking to do on this one.  It seems like we're going after a pretty broad set of use cases, some of which may not be as important.  

The default-deny user story resonates with me, however for the others (concerning pods as the "selected entities" not Nodes) I generally feel like having a pod exposed to the "outside world" is a rare case.  i.e even with all the mentioned ingress APIs there's usually a cluster -entity which routes/load balances or directs external traffic do destination pods, which we can already block traffic from today. 

Lastly we need to think carefully about having a NPEP which references multi-cluster scenarios...as I think the scope could quickly explode.

npep/npep-127.md
```diff
@@ -0,0 +1,88 @@
...
+
+## TLDR
+
+This NPEP proposes adding support for cluster ingress (southbound) traffic control in the `AdminNetworkPolicy` and `BaselineAdminNetworkPolicy` API objects.
+
+## Goals
+
+* Implement ingress traffic control from external destinations (outside the cluster)
+
+## Non-Goals
```
astoycos, line 14: Determining the non-goals here is going to be really important... I think we need to narrow this NPEP down

npinaeva: that would be nice, but I just don't have any ideas on what to put here for now. I hoped that something will be outlined in the discussion here, if you have any ideas for non-goals, please lmk :)

npep/npep-127.md
```diff
@@ -0,0 +1,88 @@
...
+(either with k8s APIs like [Ingress API](https://kubernetes.io/docs/concepts/services-networking/ingress/) or
+[Gateway API](https://gateway-api.sigs.k8s.io/), or with downstream implementations,
+like [openshift Route](https://docs.openshift.com/container-platform/4.13/rest_api/network_apis/route-route-openshift-io-v1.html)).
+Some of the APIs may provide functionality to filter incoming connections, but Network Policy API is different, because
+it is applied to the cluster workloads, and should provide the required level of security regardless of how/if a cluster workload
+is exposed to the outer world.
+
+### Use cases
+ - zero trust policy: As a cluster administrator I want to set up default deny all ingress policy as a BANP.
+
```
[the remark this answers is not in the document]
npinaeva: I am trying to distinguish ANP vs BANP policies with the word "default", because BANP is just the default overridable policy, and I am not sure is `guardrails` word makes this semantics any more obvious, wdyt?

npep/npep-127.md
```diff
@@ -0,0 +1,88 @@
...
+for TCP ports 21, 23, 161, etc.
+
+ - compromised node protection: As a cluster administrator I want to make sure a compromised node can't send traffic to the 
+cluster workloads scheduled on the other nodes.
+
+    Example: my cluster has a set of nodes with very sensitive workloads, I want to make sure that if a worker node A is
+compromised, it will not be able to affect other worker nodes. To do so, I want to create a cluster-wide policy to
+deny ingress connections from some of the cluster nodes. I may also need to explicitly allow access from the control plane nodes.
+
+ - inter-cluster communication: As a cluster manager for multiple clusters I want to make sure ingress connections
```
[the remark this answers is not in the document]
npinaeva: yeah right, I think inter-cluster part should go to the example

npep/npep-127.md
npinaeva: not sure what is the best place to clarify this, but I refer to external endpoints in this NPEP as anything outside the cluster. So I would say any IP, unless you see any problem with that?

npinaeva (commented):
thanks a lot for a review @astoycos !

npep/npep-127.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Use cases
+Most of the use cases refer to ANP priority, meaning the policy should be non-overridable by namespace owners, unless
+other clarification are mentioned in the use case itself.
+
+ - zero trust policy: As a cluster administrator I want to set up default deny all ingress policy as a BANP.
+
+    Example: when implementing a zero trust policy I want to make sure no ingress connection (unless explicitly allowed) will
+get to the cluster workloads. This should apply not only to the connections coming from the other cluster workloads, but to
+all incoming connections.
```
danwinship, line 41: I'd just add this example to the end of the use case. ", including traffic from outside the cluster"

npep/npep-127.md
```diff
@@ -0,0 +1,91 @@
...
+
+ - block well-known ports: As a cluster administrator I want to block all ingress traffic on specific ports.
+
+   Example: when defining a cluster security policy, we decided that some well-known services like ftp, telnet, SNMP, etc.
+should not be allowed for cluster workloads. To implement this policy I want to explicitly deny all ingress connections
+for TCP ports 21, 23, 161, etc.
```
danwinship, line 47: again (and in all the following cases), just put the "example" into the user story. User stories should be specific and evocative, not vague and generic.

npep/npep-127.md
```diff
@@ -0,0 +1,91 @@
...
+
+ - compromised node protection: As a cluster administrator I want to make sure a compromised node can't send traffic to the 
+cluster workloads scheduled on the other nodes.
+
+    Example: my cluster has a set of nodes with very sensitive workloads, I want to make sure that if a worker node A is
+compromised, it will not be able to affect other worker nodes. To do so, I want to create a cluster-wide policy to
+deny ingress connections from some of the cluster nodes. I may also need to explicitly allow access from the control plane nodes.
```
danwinship, line 54: Hm... so are there pods on the compromised node? Can't the attacker forge connections from pod IPs rather than the node IP? Or is it assumed that that will be blocked as well?

npep/npep-127.md
```diff
@@ -0,0 +1,91 @@
...
+
+By default, cluster workloads are not accessible from outside the cluster, but there are many ways to expose these workloads
+(either with k8s APIs like [Ingress API](https://kubernetes.io/docs/concepts/services-networking/ingress/) or
+[Gateway API](https://gateway-api.sigs.k8s.io/), or with downstream implementations,
+like [Openshift's Route CR](https://docs.openshift.com/container-platform/4.13/rest_api/network_apis/route-route-openshift-io-v1.html)).
+Some of the APIs may provide functionality to filter incoming connections, but Network Policy API is different, because
+it is applied to the cluster workloads, and should provide the required level of security regardless of how/if a cluster workload
+is exposed to the outer world.
+
+### Use cases
```
danwinship, line 33: So all of these are pretty vague about what kind of cluster ingress they're talking about, but there are a lot of different kinds, and some are harder to deal with than others, so I think it's important that our user stories explain what sorts of ingress activity it is that the admin cares about. (Even if it turns out that we declare some of these ingress types to be Non-Goals, it's good to be able to see clearly what sort of user stories we decided we can't solve right now.)

eg:
- direct delivery to hostNetwork pod
- pod HostPort
- direct delivery to pod-network pod IP (for network plugins that support that)
- service NodePort (with Cluster or Local externalTrafficPolicy)
- service ExternalIP (with Cluster or Local externalTrafficPolicy)
- service LoadBalancer IP (with Cluster or Local externalTrafficPolicy)
- direct delivery to service ClusterIP (for network plugins that support that)
- Ingress/Gateway (which generally resolves to one of the above, but exactly which one is implementation-dependent).

npep/npep-127.md
npinaeva: this user story considers that all pods from the same namespace are on the same node, and that inter-namespace communication is also denied. I omitted it since we already have the means to implement this part with namespace selectors (and potentially tenancy is you label namespace with nodes), but maybe it makes sense to add these details

npep/npep-127.md
tssurya: direct delivery to hostNetwork pod
pod HostPort

^ these seem very rare on cloud env's as use cases, can't think of any BUT on BMs can be useful but then again mostly the bastion or jump host or whatever you put on the same subnet and put accessible routes to is an entity you are consciously choosing to allow.. why would we want to block that...
```
Even if it turns out that we declare some of these ingress types to be Non-Goals, it's good to be able to see clearly what sort of user stories we decided we can't solve right now.)
```
right ^ good idea so that when we get asked why the current cidr and node peers are only egress.. we can just look at the non-goals section of this NPEP.

danwinship: > I considered the way traffic reaches the pod to be not important in this context (meaning it should work in all situations)

I mean, ideally, yes, certainly. But in practice that is impossible, because the network policy implementation is normally completely separate from cloud LBs and Gateway mechanisms and has no power over them or even any insight into how they work. Cluster ingress, as currently defined, is a very leaky abstraction.

We talked about this a little at KubeCon Paris and Shane mentioned https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-735/index.md which is a (Declined) GEP about doing policy-ish stuff in Gateway Routes. It is likely that for some user stories, providing a non-ANP policy API inside Gateway API is likely to work better than trying to implement ANP policy against arbitrary Gateway mechanisms. There was talk about doing a "GEPNPEP"...

(This is somewhat related to the problem of pod-to-cloud-LB-to-whatever traffic discussed in #203, and one thing that could potentially help to solve both would be to come up with a "bigger" model of "Kubernetes Networking" that makes the cloud network and gateways more legible to the network policy implementation. In theory, this could be part of KNI, though it is not something currently being discussed.)

PR #132, opened 2023-08-04
[WIP] Add conformance testcases for AdminNetworkPolicy named port feature

Dyanngg:
Tests can be run with `go test -SupportedFeatures AdminNetworkPolicyNamedPorts`


conformance/tests/admin-network-policy-core-egress-tcp-rules.go
```diff
@@ -61,10 +62,10 @@ var AdminNetworkPolicyEgressTCP = suite.ConformanceTest{
 			// egressRule at index0 will take precedence over egressRule at index1; thus ALLOW takes precedence over DENY since rules are ordered
 			success := kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-gryffindor", "harry-potter-0", "tcp",
 				serverPod.Status.PodIP, int32(80), s.TimeoutConfig.RequestTimeout, true)
-			assert.Equal(t, true, success)
+			assert.True(t, success)
```
tssurya, line 65: nice!
nit: split refactoring the assertion in a new commit for easier review and less distraction from main logic

conformance/tests/admin-network-policy-core-egress-tcp-rules.go
```diff
@@ -207,12 +208,65 @@ var AdminNetworkPolicyEgressTCP = suite.ConformanceTest{
...
 		})
 	},
 }
+
+var AdminNetworkPolicyEgressNamedPort = suite.ConformanceTest{
+	ShortName:   "AdminNetworkPolicyEgressNamedPort",
+	Description: "Tests support for egress traffic on a named port using admin network policy API based on a server and client model",
+	Features: []suite.SupportedFeature{
+		suite.SupportAdminNetworkPolicy,
```
tssurya, line 225: hmm shall we keep this for now under `suite.SupportAdminNetworkPolicyNamedPorts` alone? Putting it also as `suite.SupportAdminNetworkPolicy` makes it a core rather than an extended feature or should we make this a core feature?

conformance/tests/admin-network-policy-core-egress-tcp-rules.go
```diff
@@ -207,12 +208,65 @@ var AdminNetworkPolicyEgressTCP = suite.ConformanceTest{
...
 			// ensure egress from gryffindor is ALLOWED to slytherin for rest of the traffic; matches no rules hence allowed
 			success = kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-gryffindor", "harry-potter-1", "tcp",
 				serverPod.Status.PodIP, int32(8080), s.TimeoutConfig.RequestTimeout, true)
-			assert.Equal(t, true, success)
+			assert.True(t, success)
 		})
 	},
 }
+
+var AdminNetworkPolicyEgressNamedPort = suite.ConformanceTest{
```
tssurya, line 221: should we put this test and the ingress one under `test/admin-network-policy-extended-named-ports.go` ? This file is large as is, just wondering if we should mix them up like currently or keep them separate in new file for readability?

conformance/tests/admin-network-policy-core-egress-tcp-rules.go
```diff
@@ -207,12 +208,65 @@ var AdminNetworkPolicyEgressTCP = suite.ConformanceTest{
...
+			err = s.Client.Get(ctx, client.ObjectKey{
+				Name: "egress-tcp",
+			}, anp)
+			require.NoErrorf(t, err, "unable to fetch the admin network policy")
+			namedPortRule := anp.DeepCopy().Spec.Egress[5]
+			webPort := "web"
+			// replace the tcp port 8080 rule as named port rule which translate to tcp port 80 instead
+			namedPortRule.Ports = &[]v1alpha1.AdminNetworkPolicyPort{
+				{
+					NamedPort: &webPort,
```
tssurya, line 252: I am ok with doing this but just wondering if adding a rule to the existing yaml with the namedPort is better? -> It could serve as an example of how a namedPort ANP should be created for the users? I won't hold the PR for that thinking out loud on what is better moving forward.

conformance/tests/admin-network-policy-core-ingress-udp-rules.go
```diff
@@ -207,12 +208,65 @@ var AdminNetworkPolicyIngressUDP = suite.ConformanceTest{
...
+			assert.True(t, success)
 		})
 	},
 }
+
+var AdminNetworkPolicyIngressNamedPort = suite.ConformanceTest{
+	ShortName:   "AdminNetworkPolicyIngressNamedPort",
+	Description: "Tests support for ingress traffic on a named port using admin network policy API based on a server and client model",
+	Features: []suite.SupportedFeature{
+		suite.SupportAdminNetworkPolicy,
```
tssurya, line 225: same here, we probably want to decide on whether namedPorts is core or extended feature?

tssurya (changes requested):
@Dyanngg : thanks for the PR my main question is should we make namedPorts core or extended feature? Currently it is placed in the list of extended features, in which case we should remove it from the core suite of tests. Your PR description tells me its extended, and not run by default. So what I would suggest is:

1. use a new file for your new egress/ingress test and call it admin-network-policy-extended-named-ports.go (break it out of the core-egress/core-ingress files)
2. and remove suite.AdminNetworkPolicy/suite.BaselineAdminNetworkPolicy from these tests since adding them means you are including these tests in the core family?

astoycos:
poke @Dyanngg

conformance/tests/admin-network-policy-core-egress-tcp-rules.go
Dyanngg: Will remember to do this for future PRs

conformance/tests/admin-network-policy-core-egress-tcp-rules.go
Dyanngg: Yes moved them to new files

conformance/tests/admin-network-policy-core-ingress-udp-rules.go
Dyanngg: If you look at https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/suite/features.go and some test file like https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/gateway-with-attached-routes.go, it seems to me that for extended testcases we do put the core feature and the specific extended feature together in the features list

conformance/tests/admin-network-policy-core-egress-tcp-rules.go
Dyanngg: While I agree with this, I think if we decided to go down that path we should actually have a separate yaml spec for each of the policy that we use and maybe avoid hot patching the policies in the tests altogether :) in terms of named port, I don't feel it is a feature that we need special documentation/example out of.

tssurya (approved):
/lgtm

Issue #133, opened 2023-08-07
[ENHANCEMENT] Add FQDN Selectors for Egress traffic


rahulkjoshi:
**Is your enhancement request related to a problem? Please describe.**
This is an extension to https://github.com/kubernetes-sigs/network-policy-api/issues/126 which deals with the egress (northbound) use-case. Specifically, this enhancement deals with specifying [Fully-Qualified Domain Names](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) (FQDN) to identify the external peers in a connection.

**Describe the solution you'd like**
User stories will be fleshed out in the NPEP, but a rough sketch can be found [here](https://docs.google.com/document/d/1wOO6fgY0PRToJ85yC5WFDywAhwkK0m-ynLrHW-EmuGM/edit)

**Describe alternatives you've considered**
The traditional alternative is to use IP selectors to specify external peers. This can be difficult to maintain and audit. It is also not user-friendly in situations where the peer is not controlled by the policy owner and the peers IP may change over time.

**Additional context**
This is a pretty common feature already implemented by many CNIs. There has already been some discussion previously about extending Kubernetes NetworkPolicy but that was shelved due to the complexity of extending the existing v1 NetworkPolicy API ([doc](https://docs.google.com/document/d/1wOO6fgY0PRToJ85yC5WFDywAhwkK0m-ynLrHW-EmuGM/edit))

rahulkjoshi:
With the API merged, I think we can close this issue. Thanks to all who helped us deliver an awesome new selector.

Issue #126, opened 2023-07-18
[ENHANCEMENT] Add Northbound traffic support for ANP API


tssurya:
**Is your enhancement request related to a problem? Please describe.**
This is follow of up of https://github.com/kubernetes-sigs/network-policy-api/issues/28
This issue tracks adding support for northbound traffic use case and also host-networked pod traffic use case for ANP from CNI pods.

**Describe the solution you'd like**
https://github.com/kubernetes-sigs/network-policy-api/pull/117 outlines the details, it has undergone some rounds of review already at https://github.com/kubernetes-sigs/network-policy-api/pull/86 

**Describe alternatives you've considered**

**Additional context**

```[tasklist]
### Tasks
- [ ] https://github.com/kubernetes-sigs/network-policy-api/pull/117
- [ ] https://github.com/kubernetes-sigs/network-policy-api/pull/143
- [ ] Document egress samples
- [ ] https://github.com/kubernetes-sigs/network-policy-api/pull/144
- [ ] https://github.com/kubernetes-sigs/network-policy-api/pull/185
- [ ] https://github.com/kubernetes-sigs/network-policy-api/pull/193
```

tssurya:
Now that the Provisional NPEP is done, I am working on PoC to implement it and then will re-open the NPEP in the implementable phase with link to PoC.

tssurya:
Status:

1. All design phase NPEPs have merged
2. NodeSelector peer API change has merged
3. inline CIDR selector API change PR is under progress
4. Starting on conformance tests for nodeselector

tssurya:
Status:

1. All design phase NPEPs have merged
2. NodeSelector peer API change has merged
3. inline CIDR selector API change PR is merged
4. nodeSelector conformance test PR is waiting for reviews
5. starting on cidr conformance tests

tssurya:
This task is done, closing it!

PR #134, opened 2023-08-08
NPEP: FQDN Selector for Egress, User stories

rahulkjoshi:
Issue-Tracker: #133

### DIFF
--- a/npeps/npep-133.md
+++ b/npeps/npep-133.md
@@ -1,0 +1,173 @@
+# NPEP-133: FQDN Selector for Egress Traffic
+
+* Issue:
+  [#133](https://github.com/kubernetes-sigs/network-policy-api/issues/133)
+* Status: Provisional
+
+## TLDR
+
+This enhancement proposes adding a new optional selector to specify egress peers
+using [Fully Qualified Domain
+Names](https://www.wikipedia.org/wiki/Fully_qualified_domain_name) (FQDNs).
+
+## Goals
+
+* Provide a selector to specify egress peers using a Fully Qualified Domain Name
+  (for example `kubernetes.io`).
+* Support basic wildcard matching capabilities when specifying FQDNs (for
+  example `*.cloud-provider.io`)
+* Currently only `ALLOW` type rules are proposed.
+  * Safely enforcing `DENY` rules based on FQDN selectors is difficult as there
+    is no guarantee a Network Policy plugin is aware of all IPs backing a FQDN
+    policy. If a Network Policy plugin has incomplete information, it may
+    accidentally allow traffic to an IP belonging to a denied domain. This would
+    constitute a security breach.
+    
+    By contrast, `ALLOW` rules, which may also have an incomplete list of IPs,
+    would not create a security breach. In case of incomplete information, valid
+    traffic would be dropped as the plugin believes the destination IP does not
+    belong to the domain. While this is definitely undesirable, it is at least
+    not an unsafe failure.
+
+* Currently only AdminNetworkPolicy is the intended scope for this proposal.
+  * Since Kubernetes NetworkPolicy does not have a FQDN selector, adding this
+    capability to BaselineAdminNetworkPolicy could result in writing baseline
+    rules that can't be replicated by an overriding NetworkPolicy. For example,
+    if BANP allows traffic to `example.io`, but the namespace admin installs a
+    Kubernetes Network Policy, the namespace admin has no way to replicate the
+    `example.io` selector using just Kubernetes Network Policies.
+
+## Non-Goals
+
+* This enhancement does not include a FQDN selector for allowing ingress
+  traffic.
+* This enhancement only describes enhancements to the existing L4 filtering as
+  provided by AdminNetworkPolicy. It does not propose any new L7 matching or
+  filtering capabilities, like matching HTTP traffic or URL paths.
+  * This selector should not control what DNS records are resolvable from a
+    particular workload.
+* This enhancement does not provide a mechanism for selecting in-cluster
+  endpoints using FQDNs. To select Pods, Nodes, or the API Server,
+  AdminNetworkPolicy has other more specific selectors.
+  * Using the FQDN selector to refer to other Kubernetes endpoints, while not
+    explicitly disallowed, is not defined by this spec and left up to individual
+    providers. Trying to allow traffic to the following domains is NOT
+    guaranteed to work:
+    * `my-svc.my-namespace.svc.cluster.local` (the generated DNS record for a
+      Service as defined
+      [here](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#services))
+    * `pod-ip-address.my-namespace.pod.cluster.local` (the generated DNS record
+      for a Pod as defined
+      [here](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pods))
+* This enhancement does not add any new mechanisms for specifying how traffic is
+  routed to a destination (egress gateways, alternative SNAT IPs, etc). It just
+  adds a new way of specifying packets to be allowed or dropped on the normal
+  egress data path.
+* This enhancement does not require any mechanism for securing DNS resolution
+  (e.g. DNSSEC or DNS-over-TLS). Unsecured DNS requests are expected to be
+  sufficient for looking up FQDNs.
+
+## Introduction
+
+FQDN-based egress controls are a common enterprise security practice.
+Administrators often prefer to write security policies using DNS names such as
+“www.kubernetes.io” instead of capturing all the IP addresses the DNS name might
+resolve to. Keeping up with changing IP addresses is a maintenance burden, and
+hampers the readability of the network policies.
+
+## User Stories
+
+* As a cluster admin, I want to allow all Pods in the cluster to send traffic to
+  an external service specified by a well-known domain name. For example, all
+  Pods must be able to talk to `my-service.com`.
+
+* As a cluster admin, I want to allow Pods in the "monitoring" namespace to be
+  able to send traffic to a logs-sink, hosted at `logs-storage.com`
+
+* As a cluster admin, I want to allow all Pods in the cluster to send traffic to
+  any of the managed services provided by my Cloud Provider. Since the cloud
+  provider has a well known parent domain, I want to allow Pods to send traffic
+  to all sub-domains using a wild-card selector -- `*.my-cloud-provider.com`
+
+### Future User Stories
+
+These are some user stories we want to keep in mind, but due to limitations of
+the existing Network Policy API, cannot be implemented currently. The design
+goal in this case is to ensure we do not make these unimplementable down the
+line.
+
+* As a cluster admin, I want to block all cluster egress traffic by default, and
+  require namespace admins to create NetworkPolicies explicitly allowing egress
+  to the domains they need to talk to.
+
+  The Cluster admin would use a `BaselineAdminNetworkPolicy` object to switch
+  the default disposition of the cluster. Namespace admins would then use a FQDN
+  selector in the Kubernetes `NetworkPolicy` objects to allow `my-service.com`.
+  
+## API
+
+TODO: https://github.com/kubernetes-sigs/network-policy-api/issues/133
+
+## Alternatives
+
+### IP Block Selector
+
+IP blocks are an important tool for specifying Network Policies. However, they
+do not address all user needs and have a few short-comings when compared to FQDN
+selectors:
+
+* IP-based selectors can become verbose if a single logical service has numerous
+  IPs backing it.
+* IP-based selectors pose an ongoing maintenance burden for administrators, who
+  need to be aware of changing IPs.
+* IP-based selectors can result in policies that are difficult to read and
+  audit.
+
+### L4 Proxy
+
+Users can also configure a L4 Proxy (e.g. using SOCKS) to inspect their traffic
+and implement egress firewalls. They present a few trade-ofs when compared to a
+FQDN selector:
+
+* Additional configuration and maintenance burden of the proxy application
+  itself
+* Configuring new routes to direct traffic leaving the application to the L4
+  proxy.  
+
+### L7 Policy
+
+Another alternative is to provide a L7 selector, similar to the policies
+provided by Service Mesh providers. While L7 selectors can offer more
+expressivity, they often come trade-offs that are not suitable for all users:
+
+* L7 selectors necessarily support a select set of protocols. Users may be
+  using a custom protocol for application-level communication, but still want
+  the ability to specify endpoints using DNS.
+* L7 selectors often require proxies to perform deep packet inspection and
+  enforce the policies. These proxies can introduce un-desireable latencies in
+  the datapath of applications.
+
+## References
+
+* [NPEP #126](https://github.com/kubernetes-sigs/network-policy-api/issues/126):
+  Egress Control in ANP
+
+### Implementations
+
+* [Antrea](https://antrea.io/docs/main/docs/antrea-network-policy/#fqdn-based-filtering)
+* [Calico](https://docs.tigera.io/calico-enterprise/latest/network-policy/domain-based-policy)
+* [Cilium](https://docs.cilium.io/en/latest/security/policy/language/#dns-based)
+* [OpenShift](https://docs.openshift.com/container-platform/latest/networking/openshift_sdn/configuring-egress-firewall.html)
+
+The following is a best-effort breakdown of capabilities of different
+NetworkPolicy providers, as of 2023-09-25. This information may be out-of-date,
+or inaccurate.
+
+|                | Antrea                         | Calico       | Cilium       | OpenShift <br/> (current) | OpenShift <br/> (future) |
+| -------------- | ------------------------------ | ------------ | ------------ | ------------------------- | ------------------------ |
+| Implementation | DNS Snooping <br/> + Async DNS | DNS Snooping | DNS Snooping | Async DNS                 | DNS Snooping             |
+| Wildcards      | ✔                              | ️✔           | ✔            | ❌                         | ✔                        |
+| Egress Rules   | ✔                              | ️✔           | ✔            | ✔                         | ✔                        |
+| Ingress Rules  | ❌                              | ️❌           | ❌            | ❌                         | ❌                        |
+| Allow Rules    | ✔                              | ️✔           | ✔            | ✔                         | ✔                        |
+| Deny Rules     | ✔                              | ️❌(?)        | ❌            | ✔                         | ❌(?)                     |
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea/
+.vscode/
 vendor/
 site/
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -65,6 +65,7 @@ nav:
       - Provisional: 
         - npeps/npep-126-egress-traffic-control.md
         - npeps/npep-122.md
+        - npeps/npep-133.md
       - Implementable:
         - npeps/npep-137-conformance-profiles.md
         # - Experimental:


npep/npep-133.md
```diff
@@ -0,0 +1,89 @@
...
+
+* This enhancement does not include a FQDN selector for allowing ingress
+  traffic.
+* This enhancement does not include any L7 matching or filtering capabilities,
+  like matching HTTP traffic or URL paths.
+    * This selector should not control what DNS records are resolvable from a
+      particular workload.
+* This enhancement does not provide a mechanism for selecting in-cluster
+  endpoints using FQDNs. This is explicitly disallowed by the spec.
+    * To select Pods, Nodes, API Server, AdminNetworkPolicy has more first party
```
Dyanngg, line 28: which are NPEPs by themselves?

rahulkjoshi: Partially. Pods is core spec, Nodes is covered by #126. API Server is an interesting case, and so far we don't have a good selector for it yet.

npep/npep-133.md
```diff
@@ -0,0 +1,89 @@
...
+“www.kubernetes.io” instead of capturing all the IP addresses the DNS name might
+resolve to. Keeping up with changing IP addresses is a maintenance burden, and
+hampers the readability of the network policies.
+
+## User Stories
+
+* As a cluster admin, I want to allow certain workloads to send traffic to a
+  service specified by a well-known domain name (e.g. on-prem logging service)
+* As a cluster admin, I want to allow the cluster to communitcate with services
+  provided by a Cloud provider by allowing their whole domain.
```
npinaeva, line 52: should we include wildcarded dns names here too? (looks like a popular option anyway)

npep/npep-133.md
```diff
@@ -0,0 +1,89 @@
...
+expressibility, they often come trade-offs that are not suitable for all users:
+
+* L7 selectors necessarily support a select set of protocols. Customers may be
+  using a custom protocol for application-level communication, but still want
+  the ability to specify endpoints using DNS.
+* L7 selectors often require proxies to perform deep packet inspection and
+  enforce the policies. These proxies can introduce un-desireable latencies in
+  the datapath of applications.
+
+## References
```
npinaeva, line 86: maybe adding links to the existing implementations of FQDN-based selector would be useful?

npinaeva (commented):
thanks for putting this together @rahulkjoshi !

npep/npep-133.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Introduction
+
+FQDN-based egress controls are a common enterprise security practice.
+Administrators often prefer to write security policies using DNS names such as
+“www.kubernetes.io” instead of capturing all the IP addresses the DNS name might
+resolve to. Keeping up with changing IP addresses is a maintenance burden, and
+hampers the readability of the network policies.
+
+## User Stories
```
astoycos, line 47: Can you please add some explicit "real world" examples to these?

npep/npep-133.md
```diff
@@ -0,0 +1,89 @@
...
+
+* As a cluster admin, I want to allow certain workloads to send traffic to a
```
astoycos, line 49: nit:
```suggestion
* As a cluster admin, I want to allow certain workloads to send traffic to an external
```

npep/npep-133.md
astoycos: Hrm also these two use cases are pretty similar... 

For now :) this is an NPEP aiming to add FQDN selector functionality to ANP (I know it obvious for use here but we should probably add a note somewhere as well).... ANP already provides means of selecting the "cluster as a whole" OR "certain workloads in the clsuter"

I think we can narrow these user stories down to a single one and provide some explicit examples for both scenarios. 

something like "As a cluster admin, I want to apply explicit rules to my cluster which allow users to specify FQDNs as valid peers."  And then add some examples of doing so.

npep/npep-133.md
```diff
@@ -0,0 +1,89 @@
...
+* Issue:
+  [#133](https://github.com/kubernetes-sigs/network-policy-api/issues/133)
+* Status: Provisional
+
+## TLDR
+
+This enhancement proposes adding a new selector to specify egress peers using
+Fully Qualified Domain Names (FQDNs). 
+
+## Goals
```
astoycos, line 12: As stated below.... Let's be explicit... will this selector be part of ALL current Network Policy API resources?  i.e ANP and BANP?

astoycos (commented):
Good start, thanks for opening this!  This NPEP is narrow and I'm alright with that :) 

I think in the case where we may not have a bunch of proper user stories to add, but still want to specify the explicit desired use of a single user story examples can be used to add explicit color. 

i.e in this case there's really only one real desire .... add a FQDN selector to ANP + BANP however if we want to specify that the selector will include both matching by full name AND by wildcard pattern let's use examples on that single user story to do so.

npep/npep-133.md
danwinship, line 12: Need to define the expected semantics as well; everyone _wants_ the semantics to be "all outbound connections are magically matched against the FQDN rules and it just works", but we know that won't really work.

- Do we support FQDN rules for both Allow and Deny rules?
  - If we support both Allow and Deny rules, what happens if, for example, someone allows connections to `example.com` but denies connections to `example.net` (which has the same IP).
- Do we require that pods make a DNS request before trying to connect, and do we require them to make that DNS request in a particular way (eg, using the resolver pointed to by the kubelet-provided `/etc/resolv.conf`)? Or do we expect the FQDN rule to work correctly even if the pod got the IP address via some means unknown to us?
  - (This is an example of the difference between Allow vs Deny rules: if we only support FQDN Allow rules, then we can say "the FQDN rule only works if you make a proper DNS request", and we can assume users will do that because they want their traffic to be allowed. But if we support FQDN Deny rules, we can't do that, because we have to assume that attackers will "cheat" in whatever way they can to get around the limits of the implementation.)
- Do we expect/require the rule to work correctly even when the domain name in question has rapidly-changing IPs and does not necessarily return all of its IPs in response to any single DNS query? (eg, if it returns different IPs to pods in one zone than it does to pods in another zone).

npep/npep-133.md
```diff
@@ -0,0 +1,89 @@
...
+  endpoints using FQDNs. This is explicitly disallowed by the spec.
+    * To select Pods, Nodes, API Server, AdminNetworkPolicy has more first party
+      selector with better UX.
+* This enhancement does not specify the details of how traffic is routed to the
+  specified destination. For example, it does not prescribe details around NAT
+  or egress gateways.
+* This enhancement does not require a particular mechanism for securing DNS
+  resolution (e.g. DNSSEC or DNS-over-TLS).
+    * TODO: Is this up to implementations? Maybe an optional enum describing
+      what security mechanims to use and providers can chose whther to support
```
danwinship, line 36:
```suggestion
      what security mechanisms to use and providers can chose whether to support
```

danwinship, line 36: It doesn't require "particular" mechanisms for securing DNS? Does it require _any_ mechanisms?

npep/npep-133.md
```diff
@@ -0,0 +1,89 @@
...
+      or not.
+
+## Introduction
+
+FQDN-based egress controls are a common enterprise security practice.
```
danwinship, line 41: [citation required]

People _want_ this functionality, but do they already have it outside of Kubernetes? You certainly can't do it with iptables/nftables. (You can say "`iptables -d kubernetes.io -j DROP`" but that means "resolve the IP addresses of `kubernetes.io` right now and create one static rule for each IP address it returns".)

Do external firewall products implement this? If so, how _exactly_ do they implement it, since if that's what users expect, then we should just do the same thing.

npep/npep-133.md
danwinship, line 47: I think an important user story here is "As a cluster admin, I want to block all cluster egress traffic by default, and require namespace admins to create NetworkPolicies explicitly allowing egress to the domains they need to talk to."

We aren't solving that user story _right now_, but we want to make sure we remember it in the NPv2 NPEP. (And we want to make sure that we don't design ANP FQDN egress in a way that would make DNP/NPv2 FQDN egress not work.)

npep/npep-133.md
rahulkjoshi: I think there's a ton of prior art both inside and outside Kubernetes on some flavor of FQDN firewall.

- https://learn.microsoft.com/en-us/azure/firewall/fqdn-filtering-network-rules
- https://cloud.google.com/firewall/docs/firewall-policies-rule-details#domain-names-for-firewall
- https://docs.vmware.com/en/VMware-NSX/4.1/administration/GUID-63262728-CA72-47D2-8E4F-16617B63A9A4.html
- https://techdocs.f5.com/kb/en-us/products/big-ip-afm/manuals/product/network-firewall-policies-implementations-13-1-0/3.html
- https://docs.tigera.io/calico-enterprise/latest/network-policy/domain-based-policy
- https://docs.cilium.io/en/stable/security/policy/language/#dns-based
- https://docs.openshift.com/container-platform/4.13/networking/openshift_sdn/configuring-egress-firewall.html

As far as matching user expectation, I'm mostly focused on what existing K8s CNIs are providing as that's the real target demographic here. I will do some reading and definitely the goal of the NPEP is to find the least common denominator behaviors. I'm curious if you have a particular behavior you're wary of.

danwinship: OpenShift's current implementation (based on proactively resolving hostnames in FQDN rules and re-resolving when the TTL expires) has turned out to not work well at all for hosts behind load balancers/CDNs (which seem to be a lot of the hosts that people want to make rules involving).

We're probably moving to an approach where we intercept client DNS requests and update the rules based on the actual IPs returned to the client, but this isn't implemented yet, and it's unknown if it will actually work better.

npep/npep-133.md
npinaeva: > If we support both Allow and Deny rules, what happens if, for example, someone allows connections to example.com but denies connections to example.net (which has the same IP).

There is a pre-defined priority for such cases, except for one case: if the rules are listed in the same `to/from` list, which doesn't seem to have the priority semantics (yet?). But generally speaking, this problem also applies to the same pod being selected with 2 different labels, and having allow and deny rules for those. So I would say it is a valid, but probably separate problem.

> do we expect the FQDN rule to work correctly even if the pod got the IP address via some means unknown to us?

I am not sure that is possible, and would say no. As I see it, this feature is just a way to translate a dns name to a set of ips (would like to hear other opinions on this though). In that case, it makes sense to require a dns request to be sent for a dns name resolution. I imagine implementations may have more restrictions on how exactly this should happen, but I don't think we can come up with the common requirements.

> Do we expect/require the rule to work correctly even when the domain name in question has rapidly-changing IPs and does not necessarily return all of its IPs in response to any single DNS query? (eg, if it returns different IPs to pods in one zone than it does to pods in another zone).

Can this also be an implementation detail? I guess it doesn't change anything from the API point of view, and there are probably more corner-cases, e.g. replies with TTL=0 and so on.

npep/npep-133.md
danwinship: OK, it looks like:

||Azure|GCP|VMware|F5|Calico|Cilium|OpenShift (current)|OpenShift (future)|
|-|-|-|-|-|-|-|-|-|
|Implementation| DNS snooping | async DNS | DNS snooping | DNS snooping? | DNS snooping | DNS snooping | async DNS | DNS snooping |
|Wildcards?| :heavy_check_mark: | :x: | :heavy_check_mark: | :x: | :heavy_check_mark: | :heavy_check_mark: | :x: | :heavy_check_mark: |
|Egress rules? | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
Ingress rules? | :x: | :heavy_check_mark:| :x:? |:heavy_check_mark: | :x: | :x: | :x: | :x: |
Allow rules? | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
Deny rules? | :heavy_check_mark:? | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :x:? | :x: | :heavy_check_mark: | :x:? |

So...

- There seems to be a strong consensus for DNS snooping, wildcards, and "allow egress" rules.
- Most systems have "deny egress" rules.
- Most systems don't have ingress FQDN rules.
- GCP and (current) OpenShift, the only two that (definitely) use async DNS rather than DNS snooping, both explicitly warn in their docs that they don't work well with frequently-changing DNS names. None of the DNS snooping systems have such warnings.
  - GCP: "For highly dynamic domain names, there might be a discrepancy between the resolved IP addresses in the local cache and the resolved IP addresses in the firewall policy rule. In such a scenario, the egress firewall rule works reliably only when the VM uses the resolved IP addresses configured in the firewall policy rules." ... "Don't use ingress rules that use FQDN objects if the domain name resolution results are highly variable or domain name resolution uses a form of DNS-based load balancing."
  - OpenShift: "Because the egress firewall controller and pods asynchronously poll the same local name server, the pod might obtain the updated IP address before the egress controller does, which causes a race condition. Due to this current limitation, domain name usage in EgressNetworkPolicy objects is only recommended for domains with infrequent IP address changes."

npep/npep-133.md
danwinship: > > If we support both Allow and Deny rules, what happens if, for example, someone allows connections to example.com but denies connections to example.net (which has the same IP).
> 
> There is a pre-defined priority for such cases

No, priority only matters for rules that match the same things.

If I say "allow to 1.2.3.4 with priority 100" and "deny to 5.6.7.8 with priority 50" (and nothing else), then the priorities are irrelevant; all traffic to 1.2.3.4 is allowed, and all traffic to 5.6.7.8 is denied.

Likewise if I say "allow to redhat.com with priority 100" and "deny to google.com with priority 50"; all connections to redhat.com are allowed and all connections to google.com are denied.

So if I say "allow to example.com with priority 100" and "deny to example.net with priority 50", the expectation is that all traffic to example.com is allowed, and all traffic to example.net is denied, but what will _actually_ happen is that all traffic to both example.com and example.net will be denied, because the rules effectively become "allow to 93.184.216.34 with priority 100" and "deny to 93.184.216.34 with priority 50", and so the lower-precedence allow rule is overridden by the higher-precedence deny rule.

(There is really only one sane thing we can do here, which is to say "don't do that then", and a few of the implementations linked below are explicit about that. e.g., Azure notes that "FQDN filtering in application rules for HTTP/S ... can discern between two FQDNs that are resolved to the same IP address. This is not the case with FQDN filtering in network rules.")

> > do we expect the FQDN rule to work correctly even if the pod got the IP address via some means unknown to us?
> 
> I am not sure that is possible, and would say no.

In the case of not-rapidly-changing IPs, and non-wildcard, the policy implementation can resolve all the FQDNs to IPs ahead of time; this is how the existing OpenShift EgressFirewall feature works.

My point though was more that if we say "no" here, then that argues that we should _not_ support FQDN Deny rules, since I can't see any way to make them actually secure in that scenario.

> > Do we expect/require the rule to work correctly even when the domain name in question has rapidly-changing IPs
>
> Can this also be an implementation detail?

No, I don't think so; as I noted [below](https://github.com/kubernetes-sigs/network-policy-api/pull/134#discussion_r1331750773), both of the two implementations that use proactive/async DNS resolution also include warnings that they don't work well for rapidly-changing IPs, whereas none of the implementations that use DNS snooping have such a warning. So that implies that if we want to support the rapidly-changing IP use case, then are imposing certain technology choices on implementors.

danwinship: Well, OK, we _could_ leave "rapidly-changing IPs" as an implementation detail, and say that at the ANP API level, it's undefined whether they will work. But I would consider that very unsatisfying...

npinaeva: >No, priority only matters for rules that match the same things.

Exactly, that is why we have priorities: to resolve "conflicts" for the same matches. If we could guarantee the same thing is never selected by different rules, we wouldn't need priorities at all. Your example kinda shows how priorities make these overlapping rules behave somewhat predictably. The same will happen if you "allow to pod selector label1=A with priority 100" and "deny to pod selector label2=B with priority 50", and then have a pod with labels `label1=A, label2=B`. So I think adding extra docs for dns examples, and for other similar examples will be useful here.

What I wanted to stress, is that there is a way to create rules saying `"allow to example.com with priority 100" and "deny to example.net with priority 100"` by just having these rules in the same ANP peers list (and the same can happen with the existing pod selectors). In that case we will have some kind of an undefined behaviour, because there is no strict prioritization between these rules, and the same thing will be allowed and denied with the same priority.

> Well, OK, we could leave "rapidly-changing IPs" as an implementation detail, and say that at the ANP API level, it's undefined whether they will work. But I would consider that very unsatisfying... 

It is certainly nice to have pre-defined behaviour, but it may be very difficult. For example, one plugin may support rapidly-changing ips at every 5 seconds rate, and another one at every millisecond rate. Defining such details will cause "imposing certain technology choices on implementors" as you said, which doesn't sounds like a flexible API.

npep/npep-133.md
rahulkjoshi: Added a couple more details.

npep/npep-133.md
rahulkjoshi: clarified.

npep/npep-133.md
rahulkjoshi: added.

npep/npep-133.md
rahulkjoshi: I don't know that "imposing certain technology choices on implementors" is a bad thing, so long as our decision is "reasonable" and backed by legitimate use-cases. This will be an optional spec of the API, so I hope it should be ok to consulting a bunch of providers and pick a compromise.

For example, the wild-card specification makes using a polling-based solution pretty much unworkable. But there is the fact that lots of providers have made it work, and  we see user-stories that ask for this -- both result in us considering it for the spec.

* In the rapidly-changing IP case, maybe it makes sense to specify a minimum TTL we expect providers to handle correctly (maybe 5 minutes). Providers can certainly do better if their implementation is capable.
* In the allow/deny case, I haven't heard of any user stories for deny that feel super rigorous to me. I'm of the opinion that deny shouldn't be part of the spec. If providers have deny implementations, I don't think keeping it out of the spec is a huge deal (users can still rely on the implementation-specific CRs if they really need it). Happy to reverse that if we find compelling use-cases.

npep/npep-133.md
danwinship: yup, go ahead

npep/npep-133.md
```diff
@@ -0,0 +1,140 @@
...
+* Provide a selector to specify egress peers using a Fully Qualified Domain Name
+  (for example `kubernetes.io`).
+* Support a restricted set of regex matching capabilities when specifying FQDNs.
+* Currently only `ALLOW` type rules are proposed.
+  * Correctly enforcing DENY rules based on FQDN selectors is difficult as there
+    is no guarantee a Network Policy plugin is aware of all IPs backing a FQDN
+    policy. 
+* Currently only AdminNetworkPolicy is the intended scope for this proposal.
+  * Since Kubernetes NetworkPolicy does not have a FQDN selector, adding this
+    capability to BaselineAdminNetworkPolicy can result in unintended behavior.
```
danwinship, line 24:
```suggestion
  * Since Kubernetes NetworkPolicy does not have a FQDN selector, adding this
    capability to BaselineAdminNetworkPolicy would not be very useful.
```

Saying that it "can result in unintended behavior" makes it sound like there's some tricky unexpected interaction, but that's not the problem. The problem is just that all of the good user stories that would make use of FQDN BANPs only work if you have FQDN NPs too.

rahulkjoshi: Yeah, good point. Reworded

npep/npep-133.md
```diff
@@ -0,0 +1,140 @@
...
+
+This enhancement proposes adding a new optional selector to specify egress peers
+using [Fully Qualified Domain Names](https://www.wikipedia.org/wiki/Fully_qualified_domain_name)
+(FQDNs).
+
+## Goals
+
+* Provide a selector to specify egress peers using a Fully Qualified Domain Name
+  (for example `kubernetes.io`).
+* Support a restricted set of regex matching capabilities when specifying FQDNs.
```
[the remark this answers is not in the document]
rahulkjoshi: Ack, reworded

npep/npep-133.md
```diff
@@ -0,0 +1,140 @@
...
+* This enhancement does not include a FQDN selector for allowing ingress
+  traffic.
+* This enhancement does not include any L7 matching or filtering capabilities,
+  like matching HTTP traffic or URL paths.
+  * This selector should not control what DNS records are resolvable from a
+    particular workload.
+* This enhancement does not provide a mechanism for selecting in-cluster
+  endpoints using FQDNs. This is explicitly disallowed by the spec.
+  * To select Pods, Nodes, or the API Server, AdminNetworkPolicy has other more
+    specific selectors.
```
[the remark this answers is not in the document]
rahulkjoshi: Ack -- I reworded to be more inline with our discussion for Egress CIDRs.

I think leaving it undefined is the right course of action. While for Pods it's pretty trivial to expect providers to do the expected thing (it's the same as putting the Pod's IP directly in). But for something like Services things get tricky since often times providers have already load balanced Service VIPs to a Pod IP before NetPol enforcement.

We can provided better guidance and UX by using more specific selectors for K8s endpoints, so leaving this undefined seems like the best compromise. LMK what you think.

Dyanngg (commented):
LGTM overall

astoycos:
/lgtm for the user stories!

I'll let @danwinship Add the approval here when he's ready

npep/npep-133.md
```diff
@@ -0,0 +1,148 @@
...
+(FQDNs).
+
+## Goals
+
+* Provide a selector to specify egress peers using a Fully Qualified Domain Name
+  (for example `kubernetes.io`).
+* Support basic wildcard matching capabilities when specifying FQDNs (for example `*.cloud-provider.io`)
+* Currently only `ALLOW` type rules are proposed.
+  * Correctly enforcing DENY rules based on FQDN selectors is difficult as there
+    is no guarantee a Network Policy plugin is aware of all IPs backing a FQDN
```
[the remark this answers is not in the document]
rahulkjoshi: Ack, good point. I took a pass at fleshing out the argument more clearly. PTAL

npeps/npep-133.md
```diff
@@ -0,0 +1,158 @@
...
+  The Cluster admin would use a `BaselineAdminNetworkPolicy` object to switch
+  the default disposition of the cluster. Namespace admins would then use
+  a FQDN selector in the Kubernetes `NetworkPolicy` objects to allow
+  `my-service.com`.
+  
+## API
+
+TODO
+
+## Alternatives
```
danwinship, line 107: Since you mention L7 selectors below, and "changes to traffic routing" as a Non-Goal above, it might make sense to talk about L4 proxies as an Alternative here; admins could force pods to connect to external hosts through a SOCKS proxy that was configured to only accept connections to certain FQDNs. This seems like a more architecturally-correct solution to me, but users don't like it, since it requires then to change their pods to use the proxy. (And requires someone to set up the SOCKS proxy, which could potentially become a single point of failure...)

npeps/npep-133.md
```diff
@@ -0,0 +1,158 @@
...
+
+* [NPEP #126](https://github.com/kubernetes-sigs/network-policy-api/issues/126):
+  Egress Control in ANP
+
+### Implementations
+
+* [Antrea](https://antrea.io/docs/main/docs/antrea-network-policy/#fqdn-based-filtering)
+* [Calico](https://docs.tigera.io/calico-enterprise/latest/network-policy/domain-based-policy)
+* [Cilium](https://docs.cilium.io/en/latest/security/policy/language/#dns-based)
+* [Open Shift](https://docs.openshift.com/container-platform/latest/networking/openshift_sdn/configuring-egress-firewall.html)
```
danwinship, line 145: one word. "OpenShift"

rahulkjoshi: fixed.

npeps/npep-133.md
```diff
@@ -0,0 +1,158 @@
...
+    explicitly disallowed, is not defined by this spec and left up to individual
+    providers. Trying to allow traffic to the following domains is NOT
+    guaranteed to work:
+    * `my-svc.my-namespace.svc.cluster.local` (the generated DNS record for a
+      Service)
+    * `my-pod.my-namespace.svc.cluster.local` (the generated DNS record for a
+      Pod)
+* This enhancement does not specify the details of how traffic is routed to the
+  specified destination. For example, it does not prescribe details around NAT
+  or egress gateways.
```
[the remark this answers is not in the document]
rahulkjoshi: Updated the wording, let me know if that looks good.

npeps/npep-133.md
```diff
@@ -0,0 +1,158 @@
...
+  endpoints using FQDNs. To select Pods, Nodes, or the API Server,
+  AdminNetworkPolicy has other more specific selectors.
+  * Using the FQDN selector to refer to other Kubernetes endpoints, while not
+    explicitly disallowed, is not defined by this spec and left up to individual
+    providers. Trying to allow traffic to the following domains is NOT
+    guaranteed to work:
+    * `my-svc.my-namespace.svc.cluster.local` (the generated DNS record for a
+      Service)
+    * `my-pod.my-namespace.svc.cluster.local` (the generated DNS record for a
+      Pod)
```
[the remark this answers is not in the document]
rahulkjoshi: Yeah, bad copy and paste. Fixed now.

npeps/npep-133.md
```diff
@@ -0,0 +1,173 @@
...
+    * `pod-ip-address.my-namespace.pod.cluster.local` (the generated DNS record
+      for a Pod as defined
+      [here](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pods))
+* This enhancement does not add any new mechanisms for specifying how traffic is
+  routed to a destination (egress gateways, alternative SNAT IPs, etc). It just
+  adds a new way of specifying packets to be allowed or dropped on the normal
+  egress data path.
+* This enhancement does not require any mechanism for securing DNS resolution
+  (e.g. DNSSEC or DNS-over-TLS). Unsecured DNS requests are expected to be
+  sufficient for looking up FQDNs.
```
danehans, line 68: Are secure DNS requests supported by this spec?

npeps/npep-133.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Introduction
+
+FQDN-based egress controls are a common enterprise security practice.
+Administrators often prefer to write security policies using DNS names such as
+“www.kubernetes.io” instead of capturing all the IP addresses the DNS name might
+resolve to. Keeping up with changing IP addresses is a maintenance burden, and
+hampers the readability of the network policies.
```
danehans, line 76: Is the implementation of this spec expected to maintain the list of resolved IPs?

npeps/npep-133.md
```diff
@@ -0,0 +1,173 @@
...
+  require namespace admins to create NetworkPolicies explicitly allowing egress
+  to the domains they need to talk to.
+
+  The Cluster admin would use a `BaselineAdminNetworkPolicy` object to switch
+  the default disposition of the cluster. Namespace admins would then use a FQDN
+  selector in the Kubernetes `NetworkPolicy` objects to allow `my-service.com`.
+  
+## API
+
+TODO
```
danehans, line 109: Assuming this NPEP does not define the API, consider creating an issue and linking here.

npeps/npep-133.md
```diff
@@ -0,0 +1,173 @@
...
+* IP-based selectors can become verbose if a single logical service has numerous
+  IPs backing it.
+* IP-based selectors pose an ongoing maintenance burden for administrators, who
+  need to be aware of changing IPs.
+* IP-based selectors can result in policies that are difficult to read and
+  audit.
+
+### L4 Proxy
+
+Customers can also configure a L4 Proxy (e.g. using SOCKS) to inspect their
```
danehans, line 128: Consider replacing `Customers` with `Users` (repeat throughout).

npeps/npep-133.md
```diff
@@ -0,0 +1,173 @@
...
+compared to a FQDN selector:
+
+* Additional configuration and maintenance burden of the proxy application
+  itself
+* Configuring new routes to direct traffic leaving the application to the L4
+  proxy.  
+
+### L7 Policy
+
+Another alternative is to provide a true L7 selector, similar to the policies
```
danehans, line 139: Consider removing `true`.

npeps/npep-133.md
```diff
@@ -0,0 +1,173 @@
...
+provided by Service Mesh providers. While L7 selectors can offer more
+expressivity, they often come trade-offs that are not suitable for all users:
+
+* L7 selectors necessarily support a select set of protocols. Customers may be
```
danehans, line 143: I'm not following what you mean by:
>L7 selectors necessarily support a select set of protocols.

npeps/npep-133.md
```diff
@@ -0,0 +1,173 @@
...
+using [Fully Qualified Domain
+Names](https://www.wikipedia.org/wiki/Fully_qualified_domain_name) (FQDNs).
+
+## Goals
+
+* Provide a selector to specify egress peers using a Fully Qualified Domain Name
+  (for example `kubernetes.io`).
+* Support basic wildcard matching capabilities when specifying FQDNs (for
+  example `*.cloud-provider.io`)
+* Currently only `ALLOW` type rules are proposed.
```
danehans, line 19: Is all other egress denied, including in-cluster?

rahulkjoshi: Not exactly -- AdminNetworkPolicy doesn't operate on default deny semantics like that.

These rules would be paired with either a lower priority DENY rule, or use the default deny behavior of Kubernetes NetworkPolicy.

npeps/npep-133.md
rahulkjoshi: Yes that's the expectation.

npeps/npep-133.md
rahulkjoshi: Ack, it's coming in a follow-up very soon.

npeps/npep-133.md
rahulkjoshi: ack!

npeps/npep-133.md
rahulkjoshi: done!

npeps/npep-133.md
rahulkjoshi: L7 by its definition supports a certain protocol, for example it may support HTTP/1.0, but it doesn't know what to do with a user's gRPC request or something custom (if they have their own protocol for communication).

DNS Policies operate in a sort of hybrid mode where it uses DNS (which is technically L7) to get a list of IPs, but on enforcement time the policies are strictly L4 (IP and port and protocol). This gives users a lot more flexibility

npeps/npep-133.md
rahulkjoshi: For now, I think the API wants to stay out of that discussion.

So the spec won't have a field "forcing" the use of DNSSEC. But if ANP Plugins use DNSSEC to verify records, that's definitely allowed by the spec. Maybe in the future we'll add more selectors requiring the use of DNS-over-TLS for example, but that will come in future iterations.

danehans: Thanks for the additional context. I didn't expect the spec to enforce secure DNS, I just wanted to ensure that the spec doesn't prevent it. I agree that this is a use case that can be addressed in the future.

npeps/npep-133.md
danehans: Will the API(s) that support this NPEP define the expectations for managing the list?

npeps/npep-133.md
danehans: ^ definitely helps. Thanks for the clarification.

npeps/npep-133.md
danehans: @rahulkjoshi ^ is what I thought but the NPEP is not explicit in this regard. Might be worth adding a condensed form of ^ to be explicit.

danehans: >These rules would be paired with either a lower priority DENY rule...

However, the NPEP states that `DENY` rules are not supported.

astoycos:
Small poke @rahulkjoshi This seems pretty close!

npeps/npep-133.md
rahulkjoshi: We will probably do some work to define expected behaviors alongside the API. I will try to flesh those out in a follow-up as well, but as some prior art of what I'm thinking: https://docs.google.com/document/d/1wOO6fgY0PRToJ85yC5WFDywAhwkK0m-ynLrHW-EmuGM/edit#bookmark=id.61oxh3gwesyg

npeps/npep-133.md
rahulkjoshi: Yeah I can see how that's confusing -- I hope long-term it's clearer when we provide the API but, I'll try to explain and I'm open to suggestions on how to phrase this so it's easier to understand for other readers:

**Quick Note, API details will be affected by what we decide in https://github.com/kubernetes-sigs/network-policy-api/pull/144**

Roughly speaking, the new API will extend the [AdminNetworkPolicyEgressPeer](https://github.com/kubernetes-sigs/network-policy-api/blob/774376b17a86ea0a4ebbd640aedab2ec000d2816/apis/v1alpha1/shared_types.go#L153-L178) by adding a new selector. AdminNetworkPolicyEgressPeer is specified inside the [AdminNetworkPolicyEgressRule](https://github.com/kubernetes-sigs/network-policy-api/blob/774376b17a86ea0a4ebbd640aedab2ec000d2816/apis/v1alpha1/adminnetworkpolicy_types.go#L157) struct (in the repeated field `To`).

When we say `DENY` rules are not supported, more specifically what we're saying is: If a `AdminNetworkPolicyEgressRule` has `Action: "Deny"`, then none of the `AdminNetworkPolicyEgressPeer`s in that rule can specify FQDN Selectors.

You can however, specify another `AdminNetworkPolicyEgressRule` with `Action: "Deny"` (you can either make another entry in `AdminNetworkPolicySpec.Egress` or create a separate ANP object) and include other selectors in that rule.

astoycos:
/lgtm
/approve

Thanks for all the hard work @rahulkjoshi!!

astoycos (approved):
/lgtm
/approve

Thanks for all the hard work @rahulkjoshi!!!

PR #135, opened 2023-08-16
setup for v0.1.1 release

astoycos:
We're still not doing this release process 100% correct and should instead be opening a PR against the v0.1.0 branch reviewing there before merging this to update the changelog and cut the release

However, for now we're just shipping conformance fixes so that should be alright. 

Please see https://github.com/kubernetes-sigs/network-policy-api/compare/v0.1.0...v0.1.1 for all the changes since v0.1.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Table of Contents
 
 - [v0.1.0](#v010)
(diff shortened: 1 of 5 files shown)


Dyanngg (approved):
LGTM

Issue #104, opened 2023-05-21
Add pod selector matches to conformance test suite


tssurya:
Currently all conformance tests are only testing on a namespace level (users should use `NetworkPolicy` if they want to fall to a pod-level which is why the first round of tests are not giving importance to this). We need to also add tests that use `podSelector` since that is a supported part of the API.

Issue #108, opened 2023-05-23
Enhance `pokeServer` in helper utilities


tssurya:
See https://github.com/kubernetes-sigs/network-policy-api/pull/99/commits/63394cc816c1d3f78be860a1faa100a6efd4ca42#r1200872040 for details.

- [x] See if we can improve the parser to be more accurate on what an error means
- [x] See if we need to retry the connection errors similar to how its done in netpol v1 tests

/area conformance

Shin-John:
Can I still assign this to myself? I see that there has been a merged mention.

flavio-fernandes:
Here is the corrected version of your text:

A humble side note I would like to have along with this issue, which I could have used while encountering a problem during the conformance test:

```
=== RUN   TestNetworkPolicyV2Conformance/AdminNetworkPolicyEgressSCTP/Should_support_an_'allow-egress'_policy_for_SCTP_protocol;_ensure_rule_ordering_is_respected
    admin-network-policy-core-egress-sctp-rules.go:62: FAILED Command was [/agnhost connect --timeout=3s --protocol=sctp 10.192.0.3:9003]
    admin-network-policy-core-egress-sctp-rules.go:62: Expected connection to succeed from network-policy-conformance-ravenclaw/luna-lovegood-0 to 10.192.0.3, but instead it miserably failed. stderr: OTHER: cannot allocate memory
 ```

If you see OTHER: cannot allocate memory while running the test, try increasing the amount of memory on the nodes running the test pods. In my particular test setup, a 16GB VM for 3 nodes was not enough. YMMV (your mileage may vary).

tssurya:
@npinaeva will check if this issue got fixed in PR353 or not

PR #139, opened 2023-09-06
Pod selector conformance tests

olivercodes:
Add pod selector conformance tests. See https://github.com/kubernetes-sigs/network-policy-api/issues/104 for details

- [x] ANP TCP
  - ~~Ingress~~
  - ~~Egress~~
- [x] ANP UDP
  - ~~Ingress~~
  - ~~Egress~~
- [x] ANP SCTP
  - ~~Ingress~~
  - ~~Egress~~
- [x] BANP TCP
  - ~~Ingress~~
  - ~~Egress~~
- [x] BANP UDP
  - ~~Ingress~~
  - ~~Egress~~
- [x] BANP SCTP
  - ~~Ingress~~
  - ~~Egress~~

### DIFF
--- a/conformance/base/admin_network_policy/core-egress-sctp-rules.yaml
+++ b/conformance/base/admin_network_policy/core-egress-sctp-rules.yaml
@@ -70,3 +70,55 @@ spec:
         namespaceSelector:
           matchLabels:
             kubernetes.io/metadata.name: network-policy-conformance-hufflepuff
+  - name: "deny-to-professor-quirrell-everything"
+    action: "Deny"
+    to:
+    - pods:
+        namespaces:
+          namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: network-policy-conformance-hogwarts-staff
+        podSelector:
+          matchLabels:
+            statefulset.kubernetes.io/pod-name: professor-quirrell-0
+  - name: "allow-to-professor-dumbledore"
+    action: "Allow"
+    to:
+    - pods:
+        namespaces:
+          namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: network-policy-conformance-hogwarts-staff
+        podSelector:
+          matchLabels:
+            statefulset.kubernetes.io/pod-name: professor-dumbledore-0
+    ports:
+      - portNumber:
+          protocol: SCTP
+          port: 9003
+  - name: "allow-to-professor-dumbledore"
+    action: "Allow"
+    to:
+    - pods:
+        namespaces:
+          namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: network-policy-conformance-hogwarts-staff
+        podSelector:
+          matchLabels:
+            statefulset.kubernetes.io/pod-name: professor-dumbledore-0
+    ports:
+      - portNumber:
+          protocol: SCTP
+          port: 9005
+  - name: "deny-to-dumbledore-everything-else"
+    action: "Deny"
+    to:
+    - pods:
+        namespaces:
+          namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: network-policy-conformance-hogwarts-staff
+        podSelector:
+          matchLabels:
+            statefulset.kubernetes.io/pod-name: professor-dumbledore-0
--- a/conformance/base/manifests.yaml
+++ b/conformance/base/manifests.yaml
@@ -26,6 +26,13 @@ metadata:
 apiVersion: v1
 kind: Namespace
 metadata:
+  name: network-policy-conformance-hogwarts-staff
+  labels:
+    conformance-house: hogwarts-staff
+---
+apiVersion: v1
+kind: Namespace
+metadata:
   name: network-policy-conformance-ravenclaw
   labels:
     conformance-house: ravenclaw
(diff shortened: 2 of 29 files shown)


conformance/tests/admin-network-policy-core-ingress-tcp-rules.go
```diff
@@ -56,10 +56,19 @@ var AdminNetworkPolicyIngressTCP = suite.ConformanceTest{
 				Name:      "harry-potter-0",
 			}, serverPod)
 			require.NoErrorf(t, err, "unable to fetch the server pod")
+
+			// ensure ingress is ALLOWED from dumbledore and DENIED from professor quirrell. Both pods are in the hogwarts-staff namespace
```
Dyanngg, line 60: nit: Maybe mention which rules (at index 7 and 8 in this particular case) causes the traffic to be allowed/dropped

conformance/tests/admin-network-policy-core-ingress-tcp-rules.go
```diff
@@ -56,10 +56,19 @@ var AdminNetworkPolicyIngressTCP = suite.ConformanceTest{
...
+			success := kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-hogwarts-staff", "professor-dumbledore-0", "tcp",
+				serverPod.Status.PodIP, int32(80), s.TimeoutConfig.RequestTimeout, true)
+			assert.Equal(t, true, success)
+			success = kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-hogwarts-staff", "professor-quirrell-0", "tcp",
```
Dyanngg, line 64: Not related to this PR, but I think we should consider refactoring the `PokeServer` function so that it is more readable. In particular, `shouldConnect` works better as a return type rather than a parameter:
```
success, connected :=  kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-hogwarts-staff", "professor-dumbledore-0", "tcp", serverPod.Status.PodIP, int32(80), s.TimeoutConfig.RequestTimeout)
assert.True(t, success)
assert.True(t, connected) // or assert.False if expect the connection to be dropped
```
@tssurya

Dyanngg (commented):
LGTM overall

conformance/tests/admin-network-policy-core-ingress-tcp-rules.go
olivercodes: 👍🏼  Remediated in the current ones, and will continue to add index markers as I work through the rest.

conformance/tests/admin-network-policy-core-ingress-tcp-rules.go
olivercodes: It's funny you say that, because this function tripped me up the first time I ran my tests (because of said readbility), I incorrectly assumed the nature of it (assumed it it was intended to expect a false return on DENY and assert as such) and had to change my boolean values when writing the DENY examples once I looked at the function definition/params and understood it better.

So yeah 👍🏼 / +1, agree

tssurya: There is an issue I had opened that outlines the "we should refactor pokeServer" :) I admit when I initially added it, it was just a starter and deff needs enhancement in terms of readability. +1 from me as well

tssurya: https://github.com/kubernetes-sigs/network-policy-api/issues/108

astoycos:
@olivercodes Is this still a WIP or ready for more review?

olivercodes:
@astoycos still WIP - I worked on it in the mornings during my vacation the last week and have most of the policies worked out now, just finishing up the tests. I'll get this PR cleaned up and ready for a 2nd review in the next day or 2

olivercodes:
@Dyanngg @tssurya this is ready for a review now. The only thing I have left is adding comments above each test (i.e. "policy at Index XX takes effect" above each test).

tssurya:
The CI needs to be green FYI @olivercodes

conformance/base/admin_network_policy/core-egress-sctp-rules.yaml
```diff
@@ -70,3 +70,55 @@ spec:
         namespaceSelector:
           matchLabels:
             kubernetes.io/metadata.name: network-policy-conformance-hufflepuff
+  - name: "deny-to-professor-quirrell-everything"
```
tssurya, line 73: why did you have to add new rules here? Couldn't we reuse the existing rules and simply edit the yaml inside the test to add/delete/update labels?

tssurya: here and everywhere else -> I'd assume the base yamls don't have to change at all, it becomes huge if it keeps growing like this right? I'd think the test logic should be able to sort of wield in the existing rules and do what's needed.

conformance/base/manifests.yaml
```diff
@@ -25,6 +25,13 @@ metadata:
 ---
 apiVersion: v1
 kind: Namespace
+metadata:
+  name: network-policy-conformance-hogwarts-staff
+  labels:
+    conformance-house: hogwarts-staff
```
tssurya, line 31: why do we need a new namespace? IMHO its bloating up the suite, aren't 4 namespaces enough?

tssurya (changes requested):
Thanks @olivercodes for working on this!

Did you add any tests where pod labels are changed and things start/stop matching on policies due to that?
I didn't find any that involved editing pod labels to achieve things.

Also see other comments inline

Issue #136, opened 2023-08-21
[ENHANCEMENT] DeveloperNetworkPolicy: namespace-scoped part of NPv2


npinaeva:
**Is your enhancement request related to a problem? Please describe.**
I want to have the next generation of NetworkPolicies for Application Developers as an addition to Admin-focused APIs.

**Describe the solution you'd like**
I would like to create a new API object for Application Developer Network Policies, based on a new set of user stories that will be outlined in an NPEP

**Describe alternatives you've considered**
Keep using NetworkPolicy/v1 :)

**Additional context**
https://docs.google.com/document/d/10L9_ZV7nmabObFSKs4rR_t8ATtOTBPwsZKKSB6GFYp0/edit

PR #162, opened 2023-10-26
NPEP-136: Introduce DeveloperNetworkPolicy

npinaeva:
First very raw version
Tracking issue https://github.com/kubernetes-sigs/network-policy-api/issues/136


npep/npep-136.md
```diff
@@ -0,0 +1,102 @@
...
+This is not the next version of an existing NetworkPolicy API, therefore it doesn't need to be compatible with
+NetworkPolicy, but it needs to be fully interoperable (i.e. have well-defined behaviour when used with NetworkPolicy).
+
+## Introduction
+
+Mostly coming from https://docs.google.com/document/d/10L9_ZV7nmabObFSKs4rR_t8ATtOTBPwsZKKSB6GFYp0/edit
+
+We want to build a new API, starting from scratch, and not building on top of the NetworkPolicy. Therefore, this NPEP
+will define the developer-focused user stories from scratch. There are some general properties that are desired for the new API
+(that we mostly learned thanks to NetworkPolicy)
```
danwinship, line 26: note that the points below apply to ANP too...

npep/npep-136.md
```diff
@@ -0,0 +1,102 @@
...
+   between an array of length 1 and an array of length 2.
+
+DeveloperNetworkPolicy should work well together with ANP and BANP as a part of the same API.
+
+## User-Stories/Use-Cases
+
+### User Story 1: Isolate namespace
+
+As a namespace owner, I want to make sure my pods can only access other pods from the same namespace and nothing else.
+I also want to restrict incoming connections from outside the namespace.
```
danwinship, line 53: The ingress and egress cases are different stories.

You need ingress deny rules if you want to protect your pods from other pods in the cluster. You don't control other users' pods, so you can't prevent them from _trying_ to connect to your pods, so you need ingress deny rules to block their connection attempts.

But the same doesn't apply to egress deny rules. If you don't want your pod to connect to pods in other namespaces then you can just _not connect to pods in other namespaces_. The only reason you'd need an egress deny rule is if you were worried that your pods might try to connect to other namespaces even though they weren't configured to. So egress deny rules are to protect you from hackers (or bugs), not to protect you from malicious other users within the cluster like ingress deny rules.

npep/npep-136.md
```diff
@@ -0,0 +1,102 @@
...
+
+### User Story 2: Allowlist egress external endpoints, deny ingress
+
+As a namespace owner, I want to allow pods in my namespace to only access specific external endpoints 
+that I define by CIDR and/or FQDN and deny everything else, including incoming connections.
```
danwinship, line 58: A related use case that I think is important: the administrator blocks all cluster egress by default with a BANP, and so namespaces have to write DNPs with CIDR or FQDN rules to allow themselves the egress they need. (ie, the admin isnt trying to actually _block_ egress, they just want to require all namespaces to explicitly _declare_ where they are egressing to).

npep/npep-136.md
```diff
@@ -0,0 +1,102 @@
...
+
+### User Story 3: Communication between namespaces
+
+As an owner of multiple namespaces, I want to allow and deny connections between namespaces.
+I have 3 namespaces: `db`, `backend`, `server`, and I want to define the following policies:
+1. namespace `db` ony allows incoming connection from the `backend` namespace, all outgoing connection are denied.
```
danwinship, line 64:
```suggestion
1. namespace `db` only allows incoming connection from the `backend` namespace, all outgoing connection are denied.
```

npep/npep-136.md
```diff
@@ -0,0 +1,102 @@
...
+### User Story 4: Communication within a namespace
+
+As a namespace owner, I have 3 pods in my namespace: `db`, `backend`, `server`. I want to define the same policy as in Story 3,
+but inside one namespace.
+
+### User Story 5: TODO: there are still more cases to cover
+
+### Less obvious User Stories (may be better postponed to the next iteration, after we agree on the most basic cases)
+- external ingress to the namespace (requires integration with the Ingress/Gateway API)
+- using nodes as peers (struggling to figure out the use case, may be something related to hostNetwork pods)
```
danwinship, line 78: in general, end users should not need to think about nodes, except as you say, if they have hostNetwork pods

npep/npep-136.md
```diff
@@ -0,0 +1,102 @@
...
+- using services as peers (was never properly discussed before, but sounds like a valid case)
```
danwinship, line 79: Yeah, IMO ingress subjects and (in-cluster) egress peers should almost always be expressed as services rather than as pods.

The problem people always had with this is that the reverse (egress subjects / ingress peers as Services) mostly doesn't make sense, and people didn't like the asymmetry.

npep/npep-136.md
```diff
@@ -0,0 +1,102 @@
...
+(This section describes the names to be used for the feature or
+features in conformance tests and profiles.
+
+These should be `CamelCase` names that specify the feature as
+precisely as possible, and are particularly important for
+Extended features, since they may be surfaced to users.)
+
+## Alternatives
+
+We considered using NetworkPolicy v2, but compatibility constraints won't let us change it significantly.
```
danwinship, line 97: we didn't really :slightly_smiling_face: 

I think the obvious Alternative is "don't do anything, just keep using NP", but that means we don't get to improve anything

npep/npep-136.md
npinaeva: I always have this concept in mind when talking about egress namespace rules, but I think it is not written anywhere, so may be a good chance to clarify that.

npep/npep-136.md
npinaeva: we probably can consider having separate definitions for egress and ingress (we are already doing that for ANP peers, so we can also do that for subject)

PR #178, opened 2023-11-20
NPEP-122 Tenancy API: Add Tenancy API design suggestions.

npinaeva:
Tracking issue https://github.com/kubernetes-sigs/network-policy-api/issues/122

### DIFF
--- a/npeps/npep-122.md
+++ b/npeps/npep-122.md
@@ -1,7 +1,7 @@
 # NPEP-122: Tenancy API
 
 * Issue: [#122](https://github.com/kubernetes-sigs/network-policy-api/issues/122)
-* Status: Provisional
+* Status: Implementable
 
 ## TLDR
 
@@ -76,6 +76,10 @@ To make the use case more obvious we can add some details to it like the followi
     Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
     may own more than 1 Namespace.
 
+    A use case from the real user falling under this story:
+    As a cluster admin, I want to set an overridable policy to allow same namespace and deny everything else.
+    Namespaces that should be affected by this policy are selected by the presence of the "foo" label.
+
 #### Story 4.2: Create and Isolate multiple tenants in a cluster, strict isolation
 
 Strict tenancy is the "Coke vs Pepsi" sort of thing where you want each tenant to feel like it has its own cluster, 
@@ -87,12 +91,28 @@ and be totally independent of the other tenants. We can write it down like this
     from other tenants. Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant 
     may own more than 1 Namespace.
 
+    As a cluster admin, I want to be able to make expections to the strict isolation (e.g. using ANP).
+
+    - Perhaps Coke and Pepsi are considering a merger and we want to allow coke-admin-team to talk to pepsi-admin-team.
+    - Perhaps there is a sensitive database that only some tenant teams are allowed access to and I want to use ANP to 
+    gate-keep the database. OTOH, the database may be in the "database team's" tenancy for bookkeeping reasons.
+
 #### Story 4.3: Allow internal connections for tenants
+
+    BANP
+    As a cluster admin, I want to setup an overridable deny-all policy to protect namespaces by default.
+    At the same time I want to build tenants in my cluster and allow connections inside one tenant by default.
     
-    As a cluster admin, I want to build tenants in my cluster and always allow connections inside one tenant.
-    At the same time I want to setup an overridable deny-all policy to protect namespaces by default.
-    This policy should make sure internal connectivity for a tenant is always allowed, in case there are
-    lower-priority deny rules.
+    ANP
+    As a cluster admin, I want to setup a deny-all policy to only allow connections that are explicitly specified.
+    Besides allowing required cluster services (like kube-api, dns, etc.) with ANP, I want to build tenants 
+    and allow connections inside one tenant.
+
+Both user stories have zero-trust policy in mind, where every allowed connection should be explicitly specified,
+and everything else is denied. It may be set up as "by default"/overridable/BANP or strict/ANP.
+Allow connection inside one tenant in this context means skip deny rules for same tenant, and delegate same-tenant 
+policies to the namespaces NetworkPolicy. We assume that there is no reason for cluster admin to forcefully allow connections
+inside one tenant instead of delegating to NetworkPolicy.
 
 #### Story 4.4: Tenants interaction with (B)ANP
 
@@ -105,7 +125,6 @@ and be totally independent of the other tenants. We can write it down like this
 
 #### What I couldn't figure out user stories for
 
-- Skip action
 - Ports *[]AdminNetworkPolicyPort
 
 ### Existing API
@@ -158,7 +177,385 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
 
 ## API
 
-TBD
+### Tenant definition
+
+For the purposes of this NPEP we define a Tenant as a set of namespaces.
+`tenancyLabels` is a set of label keys, based on which all Namespaces affected by Tenancy are be split into Tenants.
+A Tenant is identified by values of `tenancyLabels`, which are shared by all namespaces in a given Tenant.
+
+There are 2 ways to select which namespaces should be affected by Tenancy rules:
+
+1. Use a distinct new `namespaceSelector` field to define which namespaces should be affected by Tenancy, 
+while the `tenancyLabels` field can be used to define how the selected namespaces are split into Tenants.
+
+    **Cons**: selected namespace may not have some of the `tenancyLabels`, which will likely result in introducing "None" value
+    for `tenancyLabels`.
+
+```yaml
+spec:
+  namespaceSelector:
+    matchExpression:
+       - key: system-namespace
+         operator: DoesNotExist
+       - key: user
+         operator: Exists
+  tenancyLabels:
+    - user
+```
+
+2. Overload `tenancyLabels` and implicitly apply tenancy rules only to namespaces where `tenancyLabels` are present.
+
+    **Cons**: doesn't allow the use of "kubernetes.io/metadata.name" label that may be helpful to express allow-same-namespace
+tenancy.
+
+```yaml
+spec:
+  tenancyLabels:
+    - user
+```
+
+### Peers and actions
+
+Based on the existing User Stories, Tenancy only needs action to "pass same tenant" and "deny not same tenant".
+Using both actions at the same time doesn't make a lot of sense, since "pass same tenant" action only makes sense if
+there is a (B)ANP that will deny tenancy connections. Otherwise, "deny not same tenant" is sufficient.
+
+Tenancy rules don't need to specify separate rules for ingress and egress, because
+- for `SameTenant` if ingress is denied, then egress to the same tenant may be allowed when leaving the source pod, but will
+  be denied by ingress rule when coming to the destination pod. The same applies to egress.
+- for `NotSameTenant` if ingress is denied, then egress from `Tenant1` to `Tenant2` will be allowed by `Tenant1`, but
+  considered ingress from another tenant by `Tenant2`, and denied. The same applies to egress.
+
+It means that deny in at least one direction automatically denies the other direction.
+Therefore, the only extra parameter Tenancy needs is priority/precedence, and Tenancy rules may look something like:
+
+```yaml
+spec:
+  action: PassSameTenant/DenyNotSameTenant
+```
+
+### Priorities
+
+Based on User Story 4.4, we need to have Tenancy in the same priority range as ANP and BANP. There are multiple ways to do so:
+
+We don't have a user story for multiple tenancy definition, so the ability to define more than 1 tenant can be considered
+a disadvantage (users can shoot themselves in the foot by defining multiple conflicting tenancy rules) or an advantage
+as a flexible design.
+
+1. Reuse ANP to define priority, replace other spec fields with tenancy spec.
+
+```yaml
+kind: AdminNetworkPolicy
+spec:
+  priority: 10
+  tier: Override
+  subject:
+    namespaces:
+      matchExpression:
+        - key: system-namespace
+          operator: DoesNotExist
+  # this field turns on tenancy, and turns off ingress and egress peers
+  tenancyLabels: ["user"]
+  action: PassSameTenant
+```
+
+**CONS** 
+- "switch" that enables and disables fields is not the best API design
+**UNKNOWN**
+- allows multiple tenancy definitions
+
+2. Add implicit tenancy priority to ANP with extra fields.
+
+```yaml
+kind: AdminNetworkPolicy
+spec:
+  # normal ANP part
+  subject:
+    namespaces:
+      matchExpression:
+        - key: system-namespace
+          operator: DoesNotExist
+  ingress: 
+    <...>
+  egress:
+    <...>
+  priority: 10
+  tier: Override
+  # tenancy rule with higher precedence than ingress/egress rules
+  tenancyLabels: ["user"]
+  action: PassSameTenant
+```
+
+**CONS**
+- Each (B)ANP [in/e]gress section allows [100 Rules](https://github.com/kubernetes-sigs/network-policy-api/blob/005413863450e4f97f561d7698b62d268140e2ab/apis/v1alpha1/adminnetworkpolicy_types.go#L92),
+each Rule allows [100 Peers](https://github.com/kubernetes-sigs/network-policy-api/blob/005413863450e4f97f561d7698b62d268140e2ab/apis/v1alpha1/adminnetworkpolicy_types.go#L146),
+which may be used by some implementations to "flatten" all priorities into one range, by reserving 10K priorities for each (B)ANP. 
+If we add one extra priority for Tenancy, such implementations may not have extra space between reserved ranges.
+- another priority tweak within ANP object
+  **UNKNOWN**
+- allows multiple tenancy definitions
+
+3. Create a new object with ANP priorities
+
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  priority: 10
+  tier: Override
+  subject:
+    namespaces:
+      matchExpression:
+        - key: system-namespace
+          operator: DoesNotExist
+  tenancyLabels: ["user"]
+  action: PassSameTenant
+```
+
+While multiple ANPs with the same priority are allowed, we probably can allow multiple Tenancies or Tenancy and ANP
+with the same priority, but ensuring non-overlapping priorities becomes more complicated across multiple CRDs.
+
+**CONS**: 
+- new CRD
+- makes it harder to manage priorities across CRDs
+**PROS**
+- No changes to the existing ANP object
+- Users that don't care about tenancy can just ignore this CRD
+- We can throw it away if we want to change API again
+**UNKNOWN**
+- allows multiple tenancy definitions
+ 
+4. Create a new object with implicit priorities, e.g. reserve highest-priority rule before (B)ANP
+
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  subject:
+    namespaces:
+      matchExpression:
+        - key: system-namespace
+          operator: DoesNotExist
+  tenancyLabels: ["user"]
+  tier: Override/Baseline
+  action: PassSameTenant
+```
+
+**CONS** 
+- Priorities are implicit and need to be added as extra tiers between ANP/NP/BANP.
+**PROS** 
+- No changes to the existing ANP/BANP objects
+- Users that don't care about tenancy can just ignore this CRD
+- We can throw it away if we want to change API again
+**UNKNOWN**
+- doesn't allow multiple tenancy definitions
+
+#### Example (OUTDATED)
+
+Using option 4 from the previous section on priorities specification, we can outline further details here to get yaml
+examples for every use case.
+
+For
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  tenancyLabels: ["user"]
+  precedence: ANP/BANP
+  action: PassSameTenant/DenyNotSameTenant
+```
+
+To implement user story 4.3, Tenancy rules should have higher priority than ANP/BANP.
+Considering the following priority precedence: ANP Tenancy->ANP->NP->BANP Tenancy->BANP, we can express all mentioned user stories.
+
+<details>
+<summary>Full yaml examples (with the initial fields, will be updates as we agree on the final CRD format)</summary>
+
+* 4.1 "overridable isolation"
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  tenancyLabels:
+    - "user"
+  precedence: BANP
+  action: DenyNotSameTenant
+```
+OR (second option may be more useful if there is a deny BANP in the cluster)
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  tenancyLabels:
+    - "user"
+  precedence: BANP
+  action: PassSameTenant
+---
+kind: BaselineAdminNetworkPolicy
+spec:
+  subject:
+    namespaces: {}
+  ingress:
+    - action: Deny
+      from:
+        - namespaces: {}
+  egress:
+    - action: Deny
+      to:
+        - namespaces: {}
+```
+BANP can also be replaced with deny-all BANP
+```yaml
+kind: BaselineAdminNetworkPolicy
+spec:
+  subject:
+    namespaces: {}
+  ingress:
+    - action: Deny
+      from:
+        - namespaces: {}
+  egress:
+    - action: Deny
+      to:
+        - namespaces: {}
+```
+
+* 4.2 strict isolation
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  tenancyLabels:
+    - "user"
+  precedence: ANP
+  action: DenyNotSameTenant
+```
+OR (second option may be more useful if there is a deny ANP in the cluster)
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  tenancyLabels:
+    - "user"
+  precedence: ANP
+  action: PassSameTenant
+---
+kind: AdminNetworkPolicy
+spec:
+  priority: 1
+  subject:
+    namespaces:
+      matchExpression:
+        - key: user
+          operator: Exists
+  ingress:
+    - action: Deny
+      from:
+        - namespaces: {}
+  egress:
+    - action: Deny
+      to:
+        - namespaces: {}
+```
+
+* 4.3 Allow internal connections for tenants
+BANP-level
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  tenancyLabels:
+    - "user"
+  precedence: BANP
+  action: PassSameTenant
+---
+kind: BaselineAdminNetworkPolicy
+spec:
+  subject:
+    namespaces:
+      matchExpression:
+        - key: user
+          operator: Exists
+  ingress:
+    - action: Deny
+      from:
+        - namespaces: {}
+  egress:
+    - action: Deny
+      to:
+        - namespaces: {}
+```
+
+ANP-level
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  tenancyLabels:
+    - "user"
+  precedence: ANP
+  action: PassSameTenant
+---
+kind: AdminNetworkPolicy
+spec:
+  priority: 1
+  subject:
+    namespaces:
+      matchExpression:
+        - key: user
+          operator: Exists
+  ingress:
+    - action: Deny
+      from:
+        - namespaces: {}
+  egress:
+    - action: Deny
+      to:
+        - namespaces: {}
+```
+
+* 4.4 Tenants interaction with (B)ANP
+* 4.4.1 allow from monitoring + deny from not same tenant
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  tenancyLabels:
+    - "user"
+  precedence: ANP
+  action: PassSameTenant
+---
+kind: AdminNetworkPolicy
+spec:
+  priority: 1
+  subject:
+    namespaces:
+      matchExpression:
+        - key: user
+          operator: Exists
+  ingress:
+    - action: Allow
+      from:
+        - namespaces:
+            namespaceSelector:
+              matchLabels:
+                kubernetes.io/metadata.name: monitoring-ns
+    - action: Deny
+      from:
+        - namespaces: {}
+```
+* 4.4.2 allow from same tenant + BANP deny all
+```yaml
+kind: TenancyNetworkPolicy
+spec:
+  tenancyLabels:
+    - "user"
+  precedence: BANP
+  action: PassSameTenant
+---
+kind: BaselineAdminNetworkPolicy
+spec:
+  subject:
+    namespaces:
+      matchExpression:
+        - key: user
+          operator: Exists
+  ingress:
+    - action: Deny
+      from:
+        - namespaces: {}
+```
+</details>
 
 ## Conformance Details
 


npeps/npep-122.md
```diff
@@ -158,7 +158,147 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+    be limited/protected by Tenancy rules until it gets `tenancyLabels` assigned, may potentially be a security problem.
+    
+    Example: by setting deny-from-other-tenants I want to ensure only namespaces with the same label `security-zone` value
+    can send incoming connections. That policy will be violated if a new tenant is created without `security-zone` label,
+    meaning it is not a part of any Tenant. This may potentially be solved by using allow-from-same-tenant + deny ALL
+    ANP, but it is stricter, since it affects all incoming connections, not just from other pods.
+
+### Peers and actions
+
+Based on the existing User Stories, Tenancy only needs Allow and Deny actions, and the only 2 types of peers are
```
Dyanngg, line 189: I would disagree. I think user story 4.2 indicates that there is a need for PASS in case of strict tenant isolation, meaning "as a cluster admin I want to strictly disallow cross-tenant traffic, but owners of individual tenant namespaces can decide what they want to allow/deny in the scope of these namespaces"

npinaeva: I am not sure I understand it correctly. So "strictly disallow cross-tenant traffic" means `Deny from/to notSameTenant` right? How would you combine it with PASS? maybe a yaml example would help me figure it out

Dyanngg: What I mean is, instead of "deny from/to not same tenant", admin might want to define policies in the lines of "pass from/to same tenant, and deny everything else", especially when the egress story comes into play

npinaeva: I think we need a new use case for that, the semantics is "as a cluster admin, I want to let namespaces decide what is allowed within their tenant, but strictly deny everything else" and it doesn't match any of the existing use stories
it also can be expressed as "deny from not same tenant" + "ANP PASS empty pod selector", but it is less obvious for a given use case
 @danwinship wdyt?

astoycos (changes requested):
Good start thanks @npinaeva Sorry for all the comments! 

Currently the way you intertwined the challenges, examples, and actual proposed yamls is a bit confusing for me. 
 
I think you can define all of the problems upfront (Tenancy selection/ peers and actions/ and priorities) into a single concise blurb and then reference those problems in the "possible designs, example yaml section" since IMO theres 2 general design paths going forward:

1. Allow Admins to use ANP + BANP to define tenancy
2. Provide new resources to Admins for building tenancy 

And all the other design problems (Tenancy selection/ peers and actions/ and priorities) seem to be subcategories to that initial decision.

npeps/npep-122.md
danwinship: Tenancy-based rules are inherently east-west; you're always saying "traffic is allowed/denied between namespaces where these labels have the same/different values". You don't need to use a tenancy-based rule for N/S traffic, because the outside world can never be SameTenant.

IOW, if you want to implement "traffic is allowed within tenants and denied to/from everywhere else (including N/S)", then that would be:
1. a tenancy-based rule saying "traffic between tenants is denied"
2. a non-tenancy-based rule saying "N/S traffic to/from user namespaces is denied"

in which case there's still no need for a tenancy-based pass action.

----

Taking a step back, there are two concepts that we have sort of been mixing together:
1. "tenancy" - The idea that namespaces are grouped into tenants, and that you can use "same tenant" and "not same tenant" as selectors in policies.
2. "cluster default policy" - The idea that the administrator has some high-level plan for how networking works "normally" in the cluster (ie, when there aren't more-specific policies overriding it). eg:
    - "Traffic is permitted by default."
    - "Traffic within user namespaces is permitted by default, but traffic between user namespaces is blocked." 
    - "Traffic within the cluster is allowed by default, but cluster egress traffic is blocked." 

The concept of "cluster default policy" does not explicitly exist in ANP, but it's sort of lurking in the background, and all of the user stories for tenancy involve defining part of "cluster default policy", but tenancy isn't _all_ of "cluster default policy".

Part of the "tenancy is confusing / too powerful" problem that we're fixing now was because there was nothing stopping you from using tenancy-based rules in arbitrary ANPs/BANPs, even though the user stories only required tenancy-based rules in "cluster default policy" ANPs/BANPs... The new proposals below are better, since you can only define tenancy rules that apply to everyone, rather than randomly using a sameLabels rule inside a policy that only matches some traffic.

Dyanngg: > IOW, if you want to implement "traffic is allowed within tenants and denied to/from everywhere else (including N/S)"

Well my point is, I was thinking the usecase as "I don't care about the traffic within tenants, but I want to make sure traffic is denied to/from everywhere else (including N/S)". This also is not a "cluster default policy" IMO because inside a tenant I do not want to define any rule, rather than "Traffic within user namespaces is permitted by default" which sounded like a BANP.

astoycos:
Since this NPEP is now "implementable" let's make sure Sig-Network API approvers have a look as well.

/assign @thockin @khenidak

npinaeva:
> Going back to what I was saying above about "cluster default policy", what about:

I find the fact that this struct introduces invisible priorities the most confusing part, because every `ClusterDefaultPolicyRules` Action is applied somewhere between ANP, NP and BANP (and priority is implied by action type) but also it has "which was not matched by a previous rule" semantics, which is even more confusing. (but may we can make it somehow more explicit)

>"I don't care about the traffic within tenants, but I want to make sure traffic is denied to/from everywhere else (including N/S)" is
```kind: ClusterDefaultPolicy
name: default
spec:
    tenancyLabels:
        - "user"
    ingress:
        withinTenant: DefaultAllow
        default: DefaultDeny
    egress:
        withinTenant: DefaultAllow
        default: DefaultDeny```

Don't we need `Deny` instead of `DefaultDeny` here? I thought this use case wants strict isolation for N/S

I also didn't quite get the usage/meaning of `UserNamespaces` since there were no examples.

File: npeps/npep-122.md [hunk 2]
```diff
@@ -158,7 +158,147 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
 
 ## API
 
-TBD
+### Tenant definition
+
+We define Tenant as a set of namespaces.
+All Namespaces affected by Tenancy will be split into Tenants based on their label(s), we call these labels `tenancyLabels`.
```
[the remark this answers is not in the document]
npinaeva: I will update it a little, but the point is more about the semantics of tenancyLabels here than on API fields

npinaeva:
@danwinship @astoycos @Dyanngg I tried to sum up provided comments/ideas in a new section called "Another approach". Please take another look and let me know what you think

Dyanngg:
> I find the fact that this struct introduces invisible priorities the most confusing part, because every `ClusterDefaultPolicyRules` Action is applied somewhere between ANP, NP and BANP (and priority is implied by action type) but also it has "which was not matched by a previous rule" semantics, which is even more confusing. (but may we can make it somehow more explicit)

Huge +1 here. If we are going to introduce new CRD types here (which seems like we are), interoperation between tenancy related objects and AdminNetworkPolicy objects are key here. Any tenancy rules should have a clear and explicit priority so that when there are ANPs on top of these tenancy rules, which would take effect.

npeps/npep-122.md
```diff
@@ -158,7 +158,336 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+From the user stories (and common sense) looks like the only useful combinations are `SameTenant: Allow/Skip` and `NotSameTenant: Deny`.
+We can always replace `NotSameTenant: Deny` with Tenancy using `SameTenant: Allow/Skip` with lower priority 
+and (B)ANP `Deny` policy for the same tenancy selector (or maybe even wider selector, e.g. all user pods).
+This helps focus on which policy is defined inside one tenant, and then "everything else"/"other tenants"/etc
+may be defined with existing CRDs.
+
+In this case, we only need to define tenancy for the same tenant (which seems to be easier to understand).
+Tenancy itself only needs a set of `tenancyLabels` and one of the `Allow` or `Skip` actions in every direction.
+
+If we consider tenancy to be always the highest-priority rule on ANP or BANP level, we could add one more field
```
Dyanngg, line 335: Was this agreed in the user stories? I find it quite restrictive

npeps/npep-122.md
```diff
@@ -158,7 +158,336 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
 
 ## API
 
-TBD
+### Tenant definition
+
+For the purposes of this NPEP we define a Tenant as a set of namespaces.
+`tenancyLabels` is a set of label keys, based on which all Namespaces affected by Tenancy are be split into Tenants.
```
astoycos, line 164:
```suggestion
`tenancyLabels` is a set of label keys, which are used to split selected Namespaces into Tenants.
```

npeps/npep-122.md
```diff
@@ -158,7 +158,336 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+while the `tenancyLabels` field can be used to define how the selected namespaces are split into Tenants.
+
+    **Cons**: selected namespace may not have some of the `tenancyLabels`, which will likely result in introducing "None" value
+    for `tenancyLabels`. To only select namespaces with existing labels extra `key: X, operator: Exists` condition should
+    be added to the selector.
+
+```yaml
+kind: NetworkTenancy
+spec:
+  matchExpression:
```
astoycos, line 179: Is there supposed to be a `namespaceSelector` in front of this?

astoycos (commented):
So theres still a few comments left over from the last time I reviewed that I think are relevant. 

Additionally I am still not on board with creating a new `NetworkTenancy` object..... (And I think generally most of the others are of the same opinion)

At the base of this WE need an easy way to apply rules to a set of namespaces with the ability to change priority of those rules.  ANP + BANP already provide this, if the existing semantics of the API don't work let's add a new subject type AND/OR new peer types if needed. 

Finally the "we have to define everything as being in a tenant" just doesn't seem like an issue to me, if a tenant is defined, and is blocked off from "all other tenants" that also inherently means it's blocked from ALL OTHER PODs IMO.  There doesn't need to be an explicit "NONE" tenant or anything like that.

Remember we're shooting for perfection here, but generally we need the least complex API that is agreeable with the majority of the community :)

tssurya:
> Additionally I am still not on board with creating a new `NetworkTenancy` object..... (And I think generally most of the others are of the same opinion)
> 

 +1 to this from my side

danwinship:
> I am still not on board with creating a new `NetworkTenancy` object..... (And I think generally most of the others are of the same opinion)

I agree we should not have tenant-specific objects that mirror a large percentage of the functionality of ANP/BANP. But having an object to just define what tenancy means, which would then be referenced by ANPs/BANPs, would make sense. (Individual ANPs don't define which pods are in which namespaces themselves. Why should they define which namespaces are in which tenants themselves?) Or alternatively, having [an object with a very restricted set of tenant policy-setting functionality](https://github.com/kubernetes-sigs/network-policy-api/pull/178#issuecomment-1847426288) could make sense.

> At the base of this we need an easy way to apply rules to a set of namespaces with the ability to change priority of those rules.

Our user stories do not require the ability to set _arbitrary_ priorities on tenancy-related rules.

> ANP + BANP already provide this, if the existing semantics of the API don't work let's add a new subject type AND/OR new peer types if needed.

We already tried putting tenancy into the subject/peer framework, and we didn't like how it came out. Any tenancy system that does not involve any new object types seems like it is going to end up being basically equivalent to what we had in the original API with `SameLabels`/`NotSameLabels`. It will be a weird peer type that doesn't behave like any of the other peer types, and it will have the problem that different policies can have different definitions of tenancy, even though that makes the implementation more complicated and there are no user stories for that functionality.

danwinship:
> and it will have the problem that different policies can have different definitions of tenancy, even though that makes the implementation more complicated and there are no user stories for that functionality.

In particular... this NPEP started when Nadia pointed out that `SameLabels`/`NotSameLabels` tend to result in an explosion of "implementation rules" (i.e. OVN ACLs, iptables rules, etc) because you need to turn the tenancy-based ANP rule into a set of namespace-based implementation rules.

But if we say that there is only a single definition of tenancy for the entire cluster, then that makes it much easier to represent tenancy as a concept within the implementation rules, allowing for more-optimized tenancy matches. Instead of changing "subject and peer are in the same tenant" into "subject's namespace is A or B and peer's namespace is A or B; _or_ subject's namespace is C and peer's namespace is C; _or_ ...", you can just represent it as "subject's tenant == peer's tenant".

To give two examples pertaining to non-existent ANP implementations ( :smile: )
- In openshift-sdn, k8s Namespaces all have numeric IDs associated with them, and for pod-to-pod traffic, the namespace ID of the source pod is kept in `reg0` and the ID of the destination pod is in `reg1`, so that namespaceSelector-based NetworkPolicies can be implemented by OVS flows that just check `reg0` and `reg1` rather than needing to be based on pod IPs. The same approach could be used with "tenant IDs". (In fact, the "namespace IDs" originally _were_ "tenant IDs", in openshift-sdn's "multi-tenant" mode.)
- In an nftables-based implementation, if there was only a single definition of tenancy, we could just create a set containing pairs of pod IP addresses for all such pairs of IPs that are in the same tenant. Then a tenancy-matching peer just turns into a single lookup of `{ ip saddr . ip daddr }` in the tenancy set.

(OK, so actually the nftables example could be done even with "every policy creates its own definition of tenancy", by just creating a separate set for each `SameLabels`/`NotSameLabels` expression. But you might be less inclined to implement it that way, because those sets would be _big_. The openshift-sdn example really requires a single cluster-wide definition of tenancy though, since there are only so many registers, so you can't represent "this pod is part of tenant 12 with respect to ANP X, and part of tenant 378 with respect to ANP Y, and ...")

npinaeva:
Let me try to address the main point here: using (B)ANP rules for tenancy. 
Current (B)ANP semantics is: `subject` selects pod to which all `ingress` and `egress` rules are applied.
Tenancy can't be used in this semantics, because it has its own "subject" defined by the tenancy labels.
It doesn't make a lot of sense to allow something like
```yaml
apiVersion: policy.networking.k8s.io/v1alpha1
kind: AdminNetworkPolicy
spec:
  priority: 10
  subject:
    namespaces:
      matchLabels:
        kubernetes.io/metadata.name: sensitive-ns
  ingress:
    - action: Deny
      from:
      - namespaces:
         namespaceSelector: {}
   - action: Allow
     tenantLabels: ["user"]
```
because tenancy doesn't have anything to do with the given `subject`, it just tries to be injected into existing priority system.
Therefore, we can only use tenancy in (B)ANP at the same level as `subject`.

```yaml
kind: AdminNetworkPolicy
spec:
  priority: 10
  # EITHER
  subject: <...>
  ingress: <...>
  egress: <...>
  # OR
  tenant:
# example contents of tenant, the format is not defined yet
    tenancyLabels:
      - "user"
    ingress: Allow
    egress: Allow
```
This brings a questions about what is the priority between `tenancy` and `ingress`/`egress`. We can either define it (say tenancy is always a higher prio) or only allow setting one part (see EITHER OR). I prefer the first option, but it will require re-prioritization from the existing implementations (since we need to reserve 1 extra priority before ingress and egress rules that wasn't needed before)
If you have other API options that make sense, please let me know, because ^ is the best I can do without adding a new object.

Dyanngg:
I'd personally vote for `External tenant definition, arbitrary rules in ANPs/BANPs` as I don't feel like tenancy definition should be overloaded with priority. 
On top of that, I think the above proposal still does not solve the original issue we had with sameLabels: the namespaces selected in the subject field can 1. falls exactly into tenancy definition as all of them have the tenancy label(s) 2. intersect with the tenancy definition but not all of them have the tenancy label(s) 3. disjoint from the tenancy definition. We had a hard time agreeing on what we should do to policy subjects when they don't have the tenancy labels (either no-op for these specific namespaces or group these into a "default" tenant, but none of solutions were explicit and ideal).
If we are open to tenant definition as separate object, I will propose something like:
```
kind: TenantDefinition
metadata:
  name: by-user
spec:
  labels: ["user"]
```
```
apiVersion: policy.networking.k8s.io/v1alpha1
kind: AdminNetworkPolicy
spec:
  priority: 10
  subject:
     tenantNamespaces: by-user  # select namespaces by the label list in the by-user tenant definition
  ingress:
    - action: Allow
      from:
      - tenant: Same  # 'same' and 'notSame' is well-defined here as subjects must have all the label keys
    - action: Deny
      from:
      - namespaces:
         namespaceSelector: {}
```
In this case, the only thing we would need to document is how `subject: tenantNamespaces` work, which should be pretty straightforward to explain.

Dyanngg:
@danwinship If you look at my proposal, it is actually trying to add a new field in the AdminNetworkPolicy `subject`, so that there is no chance that `namespaces who aren't part of any tenant` is selected in tenancy-rule based ANPs. Also Tenancy definitions do not have to be singleton IMO

npeps/npep-122.md
```diff
@@ -158,7 +158,147 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+spec:
+  # EITHER
+  subject: <...>
+  ingress: <...>
+  egress: <...>
+  # OR
+  tenant: <tenant reference>
+```
+
+We may want to allow either `subject`+`ingress`/`egress` or `tenant`, since having both at the same priority
```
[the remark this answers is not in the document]
Dyanngg: ^ this

npinaeva:
I think we agreed some time ago that we want to simplify tenancy subject selection as much as possible, which ended up unifying tenancy labels with subject selector (because we didn't have other use cases), so we wanted to avoid "pods in sensitive namespaces allow ingress from pods in their tenant, but not from pods in other tenants". 
We could have something like 
```
apiVersion: policy.networking.k8s.io/v1alpha1
kind: AdminNetworkPolicy
spec:
  priority: 10
  tenancySubject:
     labels: ["user"]
  ingress:
    - action: Allow
      from:
      - tenant: Same  # 'same' and 'notSame' is well-defined here as subjects must have all the label keys
    - action: Deny
      from:
      - namespaces:
         namespaceSelector: {}
```
and either make `tenancySubject` and `subject` mutually exclusive, and only allow tenancy gress rules with the former, or allow both at the same time and say that tenancy rules have their own subject. Second option is closer to @danwinship suggestion, but more clear because tenancy has its own subject without extra selectors, and the first one is closer to @Dyanngg option. 
Any option has this either/or switch to turn tenancy mode on, which is a bit ugly, but may be worth it to avoid extra CRDs.
It could also be
```
kind: AdminNetworkPolicy
spec:
  subject:
# EITHER
     tenantNamespaces: 
         labels: ["user"]
# Enables sameTenant rules
# OR
    podSelector/namespaceSelector:
```
I am pretty sure if we don't want tenancy to have its own priority, then we can avoid creating an extra CRD at all

astoycos:
> I'd personally vote for External tenant definition, arbitrary rules in ANPs/BANPs as I don't feel like tenancy definition should be overloaded with priority.

I'm fairly aligned with @Dyanngg and @danwinship  here but, I also don't hate what @npinaeva Just presented, as per usual the smaller API change should be "easier" to get done, I think as long as we can express. 

Additionally as it pertains to the "pods not explicitly in a tenant suggestions" that @danwinship presented above I'd lean towards 

> 2. Some namespaces aren't part of any tenant, but for tenancy rule purposes, all such namespaces form a tenant with each other (the "nil"/"default" tenant)

IMO one of the main use cases that will be used/desired will be, "Lets build a tenant and stop it from communicating with ALL other non-same tenants/pods" and I think this option would be the easiest to parse/ understand for users wanting to build tenants.

So I would like 

```
apiVersion: policy.networking.k8s.io/v1alpha1
kind: AdminNetworkPolicy
spec:
  priority: 10
  subject:
     tenantNamespaces: by-user  # select namespaces by the label list in the by-user tenant definition
                               # in this case, this is equivalent to {namespaces where label "user" exists}
  ingress:
    - action: Allow
      from:
      - tenant: Same  # 'same' and 'notSame' is well-defined here as subjects must have all the label keys
    - action: Deny
      from:
      - namespaces:
         namespaceSelector: {}
```

To mean the same as 
```
apiVersion: policy.networking.k8s.io/v1alpha1
kind: AdminNetworkPolicy
spec:
  priority: 10
  subject:
     tenantNamespaces: by-user  # select namespaces by the label list in the by-user tenant definition
                               # in this case, this is equivalent to {namespaces where label "user" exists}
  ingress:
    - action: Allow
      from:
      - tenant: Same  # 'same' and 'notSame' is well-defined here as subjects must have all the label keys
    - action: Deny
      from:
      - tenant: NotSame
```

Otherwise I don't want to muddle up the conversation as I feel like you've all presented really great opinions and points and I think we're close/ headed in the right direction :) It seems like we're centering in on something like Dan's 

> 1. A single external tenant definition, arbitrary rules in ANPs/BANPs:

Can we try and decide on a way forward before Friday? I'd love to start working on an actual implementation PR

danwinship:
> Any option has this either/or switch to turn tenancy mode on, which is a bit ugly, but may be worth it to avoid extra CRDs.

Well, we were avoiding extra CRDs because _that_ was ugly, so if we agree that the "either/or switch" is ugly then it's possible that extra CRDs are _less_ ugly...

astoycos:
@npinaeva If we go with @danwinship's 

`1. A single external tenant definition, arbitrary rules in ANPs/BANPs:` ^^https://github.com/kubernetes-sigs/network-policy-api/pull/178#issuecomment-1930447439 

Why do we need to "inject a new priority"?  Wouldn't priority just be governed by the ANP which references the `TenantDefinition`?

npinaeva:
> @npinaeva If we go with @danwinship's
> 
> `1. A single external tenant definition, arbitrary rules in ANPs/BANPs:` ^^[#178 (comment)](https://github.com/kubernetes-sigs/network-policy-api/pull/178#issuecomment-1930447439)
> 
> Why do we need to "inject a new priority"? Wouldn't priority just be governed by the ANP which references the `TenantDefinition`?

We discussed the disadvantages of this approach in the following comments, reflected here https://github.com/kubernetes-sigs/network-policy-api/pull/178/files#diff-4bfb2f198b356812cfea399edc57d4232a6ec7cefcdc221b2fb8531a03360d87R284-R285.
It is almost the same as what we have right now, and has the same problems we are trying to solve here.

astoycos:
OK thanks for clarifying @npinaeva  :)

npinaeva:
Latest update based on the feedback from the SIG meeting on Feb 13.
Pros ans cons for every priority option are updated, check `Final suggestion` to discuss fields of the final CRD version.

astoycos (approved):
I'm happy with the suggested way forwards and think we can hash out any final design questions in the actual API change Pull Request.  Thanks for all the work @npinaeva!

npeps/npep-122.md
```diff
@@ -158,7 +158,344 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+          operator: Exists
+  ingress:
+    - action: Deny
+      from:
+        - namespaces: {}
+  egress:
+    - action: Deny
+      to:
+        - namespaces: {}
+```
```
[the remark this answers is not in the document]
npinaeva: if an admin allows namespaces to create their own network policies, then they may deny internal tenant connection by accident, and it will beak things. Then as an admin you may want to say always allow same tenant, which in turn lets namespaces specify policies for external-to-tenant traffic, but also ensures internal tenant connectivity is not broken by accident. (meaning it should prevent internal tenant connections problems both intentionally and by accident)

npeps/npep-122.md
```diff
@@ -158,7 +158,344 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
```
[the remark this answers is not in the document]
npinaeva: first of all, I am not sure between-tenant behaviour is a necessary part of the tenancy. It is not obvious why "other tenants" should be different "other namespaces". Not sure if denying traffic for a namespace outside your tenant should be based on whether that namespace is also a part of some other tenant or not. 
Not saying it shouldn't be there, but the core concept of tenancy is `sameTenant`, which is also a dynamic selector that is not possible to express in any other way. Performance-wise I would probably even expect `notSameTenant` to be represented as a lower-priority deny rule for tenancy selector (with higher-priority `sameTenant` rule), as you don't need to create dynamic `notSameTenant` sets of pods for every tenant, but can only do that for `sameTenant`.
About including full use case (both allow and deny with their priorities): the main problem I faced is integration with ANP/BANP. When you only use `sameTenant,` you need only 1 implicit priority (for every use case we have highest-ANP and highest-BANP would work). But with deny rules, you may want deny as also highest-priority (use case 4.2) or lower than allow-from-monitoring (use case 4.4.1). Considering implicit priorities are adding more complexity, trying to track where exactly each one fits both for allow and deny rules just complicates the design (e.g. I am not sure how https://github.com/kubernetes-sigs/network-policy-api/pull/178#discussion_r1504764041 this example defined tenancy priority relative to ANP priority)

npeps/npep-122.md
```diff
@@ -158,7 +158,344 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+    - action: Allow
+      from:
+        - namespaces:
+            namespaceSelector:
+              matchLabels:
+                kubernetes.io/metadata.name: monitoring-ns
+    - action: Deny
+      from:
+        - namespaces: {}
+```
```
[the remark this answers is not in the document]
npinaeva: I am not sure in the user stories context, we ever meant "all namespaces" as explicitly all namespaces in the cluster even if they are not a part of tenancy (like why would we discuss non-tenancy namespace here in the use cases). Not sure if the wording is wrong, or we have been thinking about these use cases differently this whole time. e.g. I interpret 4.4.1 as "I want to allow ingress traffic from the monitoring namespace to all (tenancy-affected) namespaces." as you can see in the yaml example. (I guess having yamls helps clarify this)
the same applies to the next comment.

npeps/npep-122.md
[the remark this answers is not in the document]
danwinship: Yeah, I don't think that's a real use case. No admin who cares about security is going to want to _prohibit_ users from using NP within their own namespaces. (And any admin who doesn't care about security isn't going to be using ANP in the first place.)

npeps/npep-122.md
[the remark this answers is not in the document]
danwinship: > first of all, I am not sure between-tenant behaviour is a necessary part of the tenancy.

Wow. OK, I guess I understand why we're having trouble agreeing on the details of the API now.

As I see it, tenancy is 100% about between-tenant behavior and 0% about anything else.

Look at the original user story:

> As a cluster admin, I want to build tenants in my cluster that are isolated from each other by default. Tenancy may be modeled as 1:1, where 1 tenant is mapped to a single Namespace, or 1:n, where a single tenant may own more than 1 Namespace.

It only talks about isolation between tenants, not anything that happens _inside_ the tenants. And the updated 4.1 says

> ... By default, cross-tenant traffic is dropped. ...

and again, neither 4.1 nor 4.2 says anything about what happens inside tenants.

The goal of tenancy is to put up walls between certain sets of namespaces. That's it. Traffic between tenants is restricted in a particular way, but all traffic that isn't between-tenant behaves exactly like it would have without the tenant definitions; you can make rules like "everybody can talk to the apiserver" or "namespaces with a user label can't talk to the internet by default", but you don't need a tenancy API to do that. You only need the tenancy API to specify between-tenant behavior.

danwinship:
OK, Nadia and I met to talk about this.

I don't really the current "Final suggestion" as proposed, because [it makes the common cases complicated](https://github.com/kubernetes-sigs/network-policy-api/pull/178#discussion_r1504711463) because you need both a Tenancy object and a separate ANP or BANP. But as Nadia points out, my version of the proposal doesn't implement all of the user stories we previously agreed upon. But if we do actually agree on all of those user stories, then the API I had proposed wasn't a good API and we should do it differently. :slightly_frowning_face: 

Part of the problem is that the existing user stories aren't entirely clear (eg, there were several places where they talk about "allowing" traffic, which I interpreted as "don't interfere with", but in Nadia's examples she interpreted as "`action: Allow`"). And, eg, [Nadia's example for 4.3](https://github.com/kubernetes-sigs/network-policy-api/pull/178/files#diff-4bfb2f198b356812cfea399edc57d4232a6ec7cefcdc221b2fb8531a03360d87R422) used an ANP-level Allow rule, which seemed bizarre to me, but it makes a lot more sense if it was supposed to be BANP-level (which is a use case that Nadia brought up several times in our call: basically, "I want default-deny for everything except within-tenant traffic"). (But even with that interpretation it's still not implementable with my proposed API.)

So maybe we need to revisit the user stories again, at least to clarify the wording. It might also help to explicitly indicate the expected results for every interesting pair of pods in each case:
- a pod connecting to a pod in the same namespace in the same tenant
- a pod connecting to a pod in a different namespace in the same tenant
- a pod connecting to a pod in a different tenant
- a pod connecting to a pod in the same namespace, which is not in a tenant
- a pod connecting to a pod in a different namespace, neither of which is a tenant
- a pod in a tenant connecting to a pod in a non-tenant namespace
- a pod in a non-tenant namespace connecting to a pod in a tenant
- a pod in a tenant connecting to a cluster-external IP
- a pod in a non-tenant namespace connecting to a cluster-external IP

In particular, I feel (though Nadia does not) that some of the user stories lead to "bad" results that probably aren't what any administrator would actually want. (eg, "I want default-deny for everything except within-tenant traffic" means that pod to pod in a tenant namespace is allowed by default, but pod to pod in a non-tenant namespace is denied by default. Nadia says they would just have to create NPs in the non-tenant namespaces in that case, but again, to me that seems like a failure of API design on our part; we should have figured out some way to give them the behavior they wanted in both tenanted and non-tenanted namespaces all at once...)

Anyway, I'm on PTO next week so feel free to merge anything you want while I'm not around to object. :slightly_smiling_face:

npeps/npep-122.md
danwinship, line 476: Looking at this again:
- If the precedence is "ANP, Tenancy, NP, BANP" then the tenancy rule is ignored, because the ANP's Deny would run first
- If the precedence is "Tenancy, ANP, NP, BANP" then the Allow-from-monitoring rule is ignored, because the Tenancy's Skip rule would bypass the ANP.

npinaeva: In this example it works as your second option, but Skip action is only applied to sameTenant namespaces, while monitoring is **not** sameTenant. So traffic from monitoring ns won't match tenancy rules (unless there is a tenant of which monitoring is a part, then monitoring will be passed as sameTenant)

npinaeva:
>we should have figured out some way to give them the behavior they wanted in both tenanted and non-tenanted namespaces all at once...
This can be done by assigning every namespace that is not a part of a tenant a unique tenancy label, so making it a one-namespace-tenant. Potentially always-allow-same-namespace policy could be expressed as tenancy with `kubernetes.io/metadata.name` tenancy label, but if it needs to be combined with another tenancy... well we need 2 tenancies then (and they would need to be allow-based not deny-based)

npeps/npep-122.md
```diff
@@ -158,7 +167,380 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+          tenant: Same
+```
+
+**CONS** complicates selection or affected namespaces, allows strange configuration, where namespaces selected by 
+`subject` have no tenancy labels. Has the same problem as the original design.
+
+3. Create 2 objects with ANP and BANP priorities (let's say NetworkTenancy and BaselineNetworkTenancy)
+
+```yaml
+kind: NetworkTenancy
```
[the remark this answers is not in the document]
npinaeva: Good point, we actually had a working name of `TenancyNetworkPolicy` (which I haven't added to the NPEP yet, but now I will), lmk if that sounds good.
About the future use cases: we are usually trying to not predict future use cases and not over-engineer APIs before it is needed. So if we need more similar functionality at some point, we probably will create a new API and some migration tool, if that makes sense.

npeps/npep-122.md
```diff
@@ -158,7 +167,380 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+spec:
+  priority: 10
+  tenancyLabels: ["user"]
+  action: SkipSameTenant
+```
+
+While multiple ANPs with the same priority are allowed, we probably can allow multiple Tenancies or Tenancy and ANP
+with the same priority, but if we decide to only allow ANP per priority, Tenancy needs to be accounted for in the same range.
+
+**CONS**: BANP doesn't have a priority, to use this method we would need to define a priority for BANP.
```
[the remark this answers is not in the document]
npinaeva: I think you are talking about option 4.1:  Create 2 new objects with implicit priorities. (I will add that one too)
The main "con" for it is that it will be 2 new CRDs with just 2 fields, but following the existing structure is the "pro", so let's discuss which one makes more sense

npeps/npep-122.md
```diff
@@ -158,7 +167,380 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+
+4. Create 1 new object with implicit priorities.
+
+`precedence` field + reserved highest-priority rule before (B)ANP
```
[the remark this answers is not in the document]
npinaeva: skip action actually is made to allow NP "override" something that ANP can't. We explain this action sometimes as "skip=delegate to namespaced network policies". It is useful when you **want** to skip some of the following ANP rules, but you don't want to make != exceptions for e.g. same-namespace traffic. Also, cluster admin doesn't always have a decision (allow or deny) for some traffic, and wants to let network policies decide that.
As per example, if I understand it correctly, you don't need to include namespaces you don't want to be affected by tenancy to the tenancy policy (not sure if there is any other need to override tenancy policy)

npeps/npep-122.md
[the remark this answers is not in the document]
tssurya: Not sure if it helps, coming here from end-user/implementor perspective. The one single ask I have been getting always is: putting wall up between tenants (tenant here being set of namespaces); the users I know are designing almost always: "allow to youself, (isSame) deny to all other tenants(isNotSame)" => I know this is covered in one of the user stories already but re-hashing that that is the single most thing the current ANP/BANP can't do well enough; I think other things like allow {} to dns and deny {} can be done using anp/banp rules unless I am missing something.

- prio1 allow to your own tenant (I think this is 4.3 but ANP level? I have not yet heard a ask for BANP level)
- prio2 allow to DNS
- prio3 allow from monitoring
- last prio deny everything to everything total lockdown

the prio1 rule expression is what we struggle with on ANP because putting {} in the subject doesn't work as it does not allow reflecting tenancy in peers, rest can be done using anp/banp rules. so really the ask is almost always an easy way to say "please allow to my own tenant" because the deny is almost always going to be the default deny banp at the bottom which takes care of putting up walls. OR if you do the {} to {} deny in ANP that also solves that problem (I can see how the deny to other tenants can be useful if we can express that explicitly)

npeps/npep-122.md
```diff
@@ -158,7 +167,403 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+    namespaces:
+      matchExpression:
+        - key: user
+          operator: Exists
+  ingress:
+    - action: Deny
+      from:
+        - namespaces: {}
+```
+</details>
```
tssurya, line 566: all the examples provided here are priceless and well thought out, thank you Nadia!!
I have a silly question:
whichever way we decide to do, If I do:
 tenancyLabels:
    - "user" (nsA:prod, nsB:staging, nsC:prod, nsD: staging, nsE:prod)
    - "color" (nsA:red, nsB:red, nsC:blue, nsD:blue, nsE:red)
   
So subject is all namespaces that have either of these tenancyLabels set right ? (not ALL?).

then the concept of `sameTenant` means any other tenant that has `Either of these 3 keys with the same value` OR `any tenant that has ALL of these 3 keys with the same value`; is it && or || ?

so who is sameTenant here for nsA? is it nsC, nsB and nsE? or only nsE ?

npinaeva: We have mostly discussed this for one label case, but I think it applies to multiple labels too (will add a section for that)
So for one label, we decided that only namespaces that have that label will be selected to avoid potential None tenant. Therefore for multiple labels, only namespaces that have all tenancy labels will be selected for the same reason. So sameTenant for nsA is namespace with user=prod AND color=red.

npeps/npep-122.md
```diff
@@ -158,7 +167,403 @@ Fourth, the ANP subject allows using pod selectors, while tenancy use cases only
...
+      operator: DoesNotExist
+    - key: user
+      operator: Exists
+  tenancyLabels:
+    - user
+```
+
+2. Overload `tenancyLabels` and implicitly apply tenancy rules only to namespaces where `tenancyLabels` are present.
+
+    **Cons**: the simplest option that covers all given use cases.
```
npinaeva, line 197: this doesn't allow using `kubernetes.io/metadata.name` label, which is a natural choice for allow-same-namespace policy, as it means all namespaces will be selected by tenancy

fasaxc:
I think the doc is a bit harsh on same/not same labels and it misses the advantages of the approach.

> First, There is no explicit definition of "tenancy" anywhere. The administrator has an idea of "tenants are defined by the user label", but that's only true because this particular ANP happens to include that particular rule, and there's no way to find the ANP(s) that defines tenancy if you don't already know what they are.

Is this a real disadvantage?  All policies are named/organised by the user right now.  They'll know where their tenancy policy is because they will have called it "tenancy-policy", or whatever makes sense to them. 

A big pro of the same/not same approach is that, if you want to understand ANP/BANP policy behaviour, you only need to look in one place, not two.  You don't need to read the ANPs and then check the tenancy policy separately and mentally combine them (or use a tool to visualise it).  You just read the ANPs in priority order to find out what they do.  It also keeps the common bits of the syntax common; in this case, it's `action: Deny` or `action: Pass`, not `isolation: Soft` (I think that'd take me back to the docs each time!) or `action: PassIfSomeCondition`, which is very different to the model that we have in ANP, where action and condition are separate.

> Second, the SameLabels/NotSameLabels selectors behave really differently from other peers, causing multiple underlying rules to be created, and the syntax doesn't make that obvious.

From the user's point of view, this is nice: the thing I want to do looks like the other thing that I know how to do. Yes, there's a potential performance headache for us but I think this is manageable between a docs warning and a little work in the impl.  Right now, users who need this use an operator or similar to auto-create a policy per namespace.  We already need to support this level of scale because users should already be able to make a policy per namespace.

In practice, I think same/not same opens up optimisation opportunities vs the old operator approach. Calico's implementation would likely look like this:

On each node:

* For each _local_ pod, add its namespace to a set.
* For each namespace in the set, generate copy of the policy with same/not same replaced by "real" selectors.  
* Dedupe policies that ended up exactly the same due to namespaces sharing labels.
* Feed results into the normal policy engine.
* ... deal with changes to policy/active set of namespaces/etc to trigger recalculation efficiently (this is SOP for us).

That actually compares pretty favourably with needing a policy per namespace just to do isolation. Debugging a recent issue for a user who had tens of thousands of namespaces, each with a generated policy, I found that:

* Storing tens of thousands of policies in RAM was causing excessive RAM usage.
* Having all those potentially-active policies was meaning that we were indexing a selector for each one in order to check whether it should apply to any local pod.

Both of those overheads would go down a lot if the user could have written a same/not same policy.  Yes, we'd expand that policy into one policy per local pod (in the worst case where they're all from different namespaces), but that's hundreds of policies instead of the tens of thousands that we had to deal with before.  It'd also allow a bit more than the tenancy use case, which would have been good for this user: their per-namespace policies also contained intra-namespace rules to lock down their application traffic.  If we only implemented tenancy isolation then **they'd still have needed a per-namespace policy** to do _that_ isolation so we're back to square one on performance.  

> Third, the syntax is very general purpose / powerful. ANP has subjects and peers, which are different, and currently Tenancy is defined on the peers side. Tenancy by itself has the same subject and peer, at least for the existing use cases, and having separate selectors for subject and peer allows for more configurations than needed.

I think this is a case where a good example in the docs that covers a simple tenancy label would head off the problem.  Most users will copy-paste _that_ and be done; only more advanced users will play with the more detailed behaviour that's possible (and they may have use cases that we haven't considered yet).  

I think same/not same is easy to grok, as evidenced by the fact that it tends to be the feature that our users ask for when they run into these policy templating/tenancy problems.  A nice aspect of it is that it covers more than just tenancy; it covers templating standard policies that you want in each namespace.  That's not a use case we've called out so far but I think it's powerful/useful.

> Fourth, the ANP subject allows using pod selectors, while tenancy use cases only need namespace selectors.

I don't see this as a problem for the user at all.  If they only want namespace isolation, I don't think they'll be motivated to add a podSelector; they'd have to go out of their way to do it.  So, yes, it's more than _needed_ but I don't think it's going to confuse anyone. Implementation-wise, it'd come "for free" with the approach that I'm picturing above (but of course other implementations may differ!)

Just to spell out "use cases that we haven't considered yet", I can think of a couple that might arise down the line where we might regret a singleton Tenancy approach:

* Multi-level isolation: perhaps you want to hard isolate Pepsi and Coke and then hard isolate pepsi-prod from pepsi-staging and then soft isolate namespaces by default but allow overrides within prod/staging.  I think we could get really stuck trying to do anything like that if we adopted a singleton approach.
* Policy templating: writing the same policy in every namespace but specialising it per namespace.  This would have been handy for the user who had 10s of thousands of policies, I think.

In any case, given that I'm convinced that we should prioritise Tenancy isolation vs ANP, I'm much more keen on approaches that embed the Pass/Drop into ANP/BANP rules rather than make an extra object with quite different semantics.  Same/NotSame feels quite natural in Calico's model; so much so that I think it's how we'd implement the other proposals under the covers(!) if it came to that.

I could probably live with the proposal that used a separate resource to name the tenancy labels, which is then used by reference in an ANP rule; I suppose we could go beyond a singleton in future if needed (or we make the tenancy labels CRD contain a list of max n labels to represent the different levels of tenency)?  The indirection adds a bit of work in our engine vs just having the labels in hand in the rule but that's not too bad.

thockin:
> I think the doc is a bit harsh on same/not same labels and it misses the advantages of the approach.

me too, but I am biased :)

>> First, There is no explicit definition of "tenancy" anywhere.

Tenancy is defined by the set of labels that "someone who knows what a tenant is" defines.

>> Second, the SameLabels/NotSameLabels selectors behave really differently from other peers, causing multiple underlying rules to be created, and the syntax doesn't make that obvious.
>
> From the user's point of view, this is nice: the thing I want to do looks like the other thing that I know how to do. Yes, there's a potential performance headache for us but I think this is manageable between a docs warning and a little work in the impl. Right now, users who need this use an operator or similar to auto-create a policy per namespace. We already need to support this level of scale because users should already be able to make a policy per namespace.

In reading the docs, this is exactly my feeling.  If there's a problem, it's not about absolute scale -- the cluster admin is going to jam the same number of "tenants" into the cluster either way.  Without a short notation, we're just forcing them to unroll the loop themselves.

I can buy that there's some work to do when we encounter a NEW such rule - run back through all known namespaces, extract this new label, and track an index.  Something like that, but I don't think it's so egregious...?

> I think same/not same is easy to grok

Again, I am biased, but I agree.  The pepsi-vs-coke analogy resonates.

> I could probably live with the proposal that used a separate resource to name the tenancy labels

I don't think we should be trying to define explicit tenancy APIs - it's going to attract a ton of attention and it will get stuck goingin in circles.

I'm not saying I LOVE the same/not-same approach, just that it seems to make sense to people.  And EVERY SINGLE customer or community person I have spoken to about tenancy could use it.  I have not seen a single approach to tenancy that doesn't use labels (or couldn't add labels to their system).

Issue #94, opened 2023-05-09
Add a Script linter to our CI workflow


astoycos:
We're maintaining a ton of scripts here, it'd be nice to have a shell script specific linter like https://www.shellcheck.net/ or something to make sure we don't miss silly things :)

v1ku:
Hello @astoycos ,

I'm working on integrating ShellCheck into the project as requested in this issue. Before proceeding, I would like to clarify a few points about the CI/CD pipeline.

    Is there an existing CI/CD pipeline for this project that I may have missed? If yes, could you please point me to the relevant configuration files?
    If there isn't a CI/CD pipeline set up, are developers expected to run the ShellCheck script manually before pushing their changes to the repository?

astoycos:
Hi @v1ku, thanks for taking this!! 

There is an existing ci/cd pipeline setup managed with prow, let me point you to a PR that should help guide you here.

I think first thing is adding a make target to our local setup to do so and then actually adding a new step to our test-infra setup.

https://github.com/kubernetes-sigs/network-policy-api/issues/37 _should_ help answer any questions.

Meanwhile feel free to ping me here or on slack for more help.

astoycos:
@v1ku Will you have time to tackle this in the near future? If not please unassign and open up for other contributors 👍

Issue #37, opened 2022-08-18
Add Gofmt and Govet gh actions 


astoycos:
Add GH actions to ensure the generated types have been created and gofmt has been run on any changes to the api itself (see the `make all` makefile target)

astoycos:
@asim-reza are you going to have time for this? I would like to unassigned this to allow other folks to take it if not :)

olivercodes:
I'll take it if it's open :)

asim-reza:
Hey! Sorry got swamped with work... please feel free to take it @olivercodes ... 
/unassign

PR #181, opened 2023-11-25
add shellcheck verification step to makefile

gabrielggg:
Hi.

This is to resolve https://github.com/kubernetes-sigs/network-policy-api/issues/94

This [install.sh](https://raw.githubusercontent.com/gabrielggg/shellcheck/master/install.sh) file is a script i created to simplify the instalation of shellcheck, it is a fork from this script https://github.com/golangci/golangci-lint/blob/master/install.sh which is used on the verify-golint.sh script from this repo. If you want i can upload that script also here, now it is hosted on my personal repo.

I took the same approach you used on https://github.com/kubernetes-sigs/network-policy-api/issues/37 creating a verify-shellcheck.sh script so that it gets automatically included on the ci pipeline because of the hack/verify-all.sh script wildcard usage.

p.s. This is my first PR to a kubernetes project, so i hope everything goes well.

### DIFF
--- a/hack/verify-shellcheck.sh
+++ b/hack/verify-shellcheck.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+# Copyright 2022 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly VERSION="v0.9.0"
+OWNER=koalaman
+REPO="shellcheck"
+BINARY=shellcheck
+FORMAT=tar.xz
+#supported OS's linux and darwin
+#supported cpu architectures: 
+#linux: x86_64, aarch64, armv6hf
+#darwin: x86_64 (only this architecture because the official shellcheck repo only delivers this binary)
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+ARCH=$(uname -m)
+GITHUB_DOWNLOAD=https://github.com/${OWNER}/${REPO}/releases/download
+NAME=${BINARY}-${VERSION}.${OS}.${ARCH}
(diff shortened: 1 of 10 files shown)


astoycos:
/ok-to-test

Thanks for your first PR!! @gabrielggg I'll review today

hack/verify-shellcheck.sh
```diff
@@ -0,0 +1,31 @@
...
+set -o nounset
+set -o pipefail
+
+readonly VERSION="v0.9.0"
+
+echo "Installing shellcheck..."
+#this install.sh file is a script i created to simplify the instalation of shellcheck
+#it is a fork from this script https://github.com/golangci/golangci-lint/blob/master/install.sh
+#which is used on the verify-golint.sh script from this repo
+curl -sfL https://raw.githubusercontent.com/gabrielggg/shellcheck/master/install.sh| sh -s "$VERSION"
```
astoycos, line 27: So I would like to avoid adding in an extra dep on your personal repository here, can we copy over just the relevant bits from `install.sh`? 

More specifically I don't think we need to support windows, we can fix the version of shellcheck being used, and we can assume that curl is there. 

It seems like we _should_ be able to just download the statically compiled binary for linux from https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.linux.x86_64.tar.xz and then run it within a script! 

With the main goal being, let's keep it as simple as possible :)

astoycos (changes requested):
Thanks for tackling this, one main comment below!

astoycos (commented):
Looks much better now @gabrielggg Can you now just fix all the scripts up in a new commit so that the new test passes? 

Also please squash these commits before doing so

gabrielggg:
hi @astoycos , i squashed the commits and then proceeded to fix all the scripts so that the test is passing now, please check it out!

gabrielggg:
Done with the rebase @astoycos

astoycos:
Additionally please try and signoff your commits :)

gabrielggg:
Hey @astoycos , i fixed the last failing tests and signed the K8s CLA, but for some reason, it still saying missing authorization on the EasyCLA item of the pipeline, can you please check that out?

astoycos:
Hrm not sure why the CLA isn't re-checking 🤔

astoycos:
@gabrielggg Thanks for the update!! I'm thinking you may need to sign your commits with the same email used for the CLA verification? 

Additionally it looks like you used different git accounts during the process and that _may_ be messing things up 

`Gomez authored and Gabriel Fernando Gomez De La Torres Parod committed` 

It should look something like 

image

gabrielggg:
hey @astoycos , thanks for the advice, i did that and know i think we are ready to go!

astoycos (approved):
/lgtm
/approve

Congrats on your first commit @gabrielggg and thanks very much!

PR #200, opened 2024-02-22
Add API spec for FQDN selector

rahulkjoshi:
Issue https://github.com/kubernetes-sigs/network-policy-api/issues/133

Provides the proposed API spec as well as some expected behaviors for both application Pods and ANP implementors to follow.


npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -106,7 +108,133 @@ line.
...
+      new connections to that IP. Existing connection will still be allowed
+      (that's consistent with NetworkPolicy behavior on long-running
+      connectionsExisting connection will still be allowed (that's consistent
+      with NetworkPolicy behavior on long-running connections).
+1. Targeting in-cluster endpoints with FQDN selector is not guaranteed to work:
+   *  Selecting Pods using their [generated DNS
+      record](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pods)
+      (for example `pod-ip-address.my-namespace.pod.cluster.local`) is not
+      guaranteed to work.
+   *  Selecting Clusters services using their [generated DNS
```
astoycos, line 234: Hrm why not?

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -106,7 +108,133 @@ line.
...
+      is not guaranteed to flow.
+   *  Pods should respect the TTL of DNS records they receive. Trying to
+      establish new connection using DNS records that are expired is not
+      guaranteed to work.
+   *  When the TTL for a DNS record expires, the implementor should stop allowing
+      new connections to that IP. Existing connection will still be allowed
+      (that's consistent with NetworkPolicy behavior on long-running
+      connectionsExisting connection will still be allowed (that's consistent
+      with NetworkPolicy behavior on long-running connections).
+1. Targeting in-cluster endpoints with FQDN selector is not guaranteed to work:
```
astoycos, line 229: Again I think this could be reallly useful for folks

astoycos (commented):
Gave this a quick pass and I'll give it another one this week thanks @rahulkjoshi

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -106,7 +108,133 @@ line.
...
+    //
+    // +optional
+    // +kubebuilder:validation:Pattern=`^([a-zA-Z0-9*]([-a-zA-Z0-9_*]*[a-zA-Z0-9*])*\.?)*$`
+    Pattern string `json:"pattern,omitempty"`
+}
+```
+
+### Examples
+
+#### Pods in `monitoring` namespace can talk to `my-service.com` and `*.cloud-provider.io`
```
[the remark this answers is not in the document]
rahulkjoshi: Added another couple examples.

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,447 @@
...
+    //      not.
+    //    - `*.*.kubernetes.io` matches subdomains of kubernetes.io at that
+    //      level.
+    //      "sub.subdomain.kubernetes.io" matches, however "kubernetes.io" and
+    //      "blog.kubernetes.io" do not.
+    //    - `*a*.kubernetes.io` matches subdomains of kubernetes.io that contain
+    //      an "a" at that level.
+    //      "api.kubernetes.io" and "domain.kubernetes.io" match, however
+    //      "blog.kubernetes.io" and "kubernetes.io" do not
+    //
```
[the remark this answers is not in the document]
rahulkjoshi: I removed the more complicated regex matching examples. They aren't really necessary, I just got a little carried away with the matcher 😄 

I've left just *.kubernetes.io and *.*.kubernetes.io. I think that captures all the use-cases. I think the user-story for *.*.kubernetes.io is basically the same as for *.kubernetes.io, but for now `*` is just scoped to one subdomain level. Happy to revisit that.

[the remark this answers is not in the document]
npinaeva: I guess we need to also change validation to not allow asterisks in random places. 
Also wildcard matching may be expensive, and allowing two-subdomain matches raises potential performance concerns, and also questions like: why only 2 subdomain and not 3? So maybe we could start with 1 wildcard match and then extend it in the future?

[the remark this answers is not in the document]
npinaeva: it was also brought to my attention, that it may be better to only allow one wildcard label and match >0 labels instead of doing `*.*.domain.com` as per https://www.rfc-editor.org/rfc/rfc1034#section-4.3.3
```
Another way to look at this is that the "*" label always matches at least one whole
label and sometimes more, but always whole labels.
```

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,438 @@
...
+        ```
+        # REQUEST 1
+
+        www.kubernetes.io  --  CNAME to kubernetes.io
+        
+        # REQUEST 2
+
+        kubernetes.io      --  A     to 1.2.3.4
+        ```
+        The implementer can still deny traffic to `1.2.3.4` because no single
```
Dyanngg, line 340: Thanks for the clarification. In our implementation I don't think we really looked into this and always assumed cname chasing

Dyanngg (approved):
LGTM

npeps/npep-133-fqdn-egress-selector.md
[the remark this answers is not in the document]
rahulkjoshi: I don't really have a strong preference one way or the other -- whether `*` matches 1 or multiple labels.

@Dyanngg @astoycos any thoughts?

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,438 @@
...
+  to all sub-domains using a wild-card selector -- `*.my-cloud-provider.com`
+
+### Future User Stories
+
+These are some user stories we want to keep in mind, but due to limitations of
+the existing Network Policy API, cannot be implemented currently. The design
+goal in this case is to ensure we do not make these unimplementable down the
+line.
+
+* As a cluster admin, I want to block all cluster egress traffic by default, and
```
astoycos, line 101: This is possible now that the egress npep has merged right, or am I mis-understanding?

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,438 @@
...
+      traffic to the IP (either because of caching or a static config), traffic
+      is not guaranteed to flow.
+   *  Pods should respect the TTL of DNS records they receive. Trying to
+      establish new connection using DNS records that are expired is not
+      guaranteed to work.
+   *  When the TTL for a DNS record expires, the implementor should stop
+      allowing new connections to that IP. Existing connection will still be
+      allowed (that's consistent with NetworkPolicy behavior on long-running
+      connections). 
+1. Targeting in-cluster endpoints with FQDN selector:
```
astoycos, line 291: Wait in the "non-goals" section above don't we say 

```
Using the FQDN selector to refer to other Kubernetes endpoints, while not
    explicitly disallowed, is not defined by this spec and left up to individual
    providers. Trying to allow traffic to the following domains is NOT
    guaranteed to work:
    * `my-svc.my-namespace.svc.cluster.local` (the generated DNS record for a
      Service as defined
      [here](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#services))
    * `pod-ip-address.my-namespace.pod.cluster.local` (the generated DNS record
      for a Pod as defined
      [here](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pods))
``` 

So why is this "supported' here?

npeps/npep-133-fqdn-egress-selector.md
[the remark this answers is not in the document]
astoycos: Hrm so I think one of these options
- Leave it up to the implementation to specify
OR
- Start with specifying that it matches only 1 label and expand in the future

(I'd probably go with the latter for now)

astoycos:
Just a few minor questions, otherwise great job @rahulkjoshi!

npeps/npep-133-fqdn-egress-selector.md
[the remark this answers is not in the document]
npinaeva: Starting simple probably makes sense, but we should add a criteria for beta-graduation, saying that we need to revisit the semantics of `*`, since changing it is not backwards-compatible.

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,438 @@
...
+
+type FQDNPeerSelector struct {
+    // Name specifies the exact Fully-Qualified Domain Name to select (the
+    // trailing '.' is optional)
+    //
+    // Support: Extended
+    //
+    // +optional
+    // +kubebuilder:validation:Pattern=`^([a-zA-Z0-9]([-a-zA-Z0-9_]*[a-zA-Z0-9])*\.?)*$`
+    Name string `json:"names,omitempty"`
```
aojea, line 138: is this not a subset of Pattern?

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,438 @@
...
+  subject:
+    namespaces: {}
+  ingress:
+    - action: Deny
+      to:
+      - networks:
+        - "0.0.0.0/0"
+```
+
+### Expected Behavior
```
aojea, line 264: do you want to define some behavior for PR records?

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,438 @@
...
+  subject:
+    namespaces:
+      matchLabels:
+        kubernetes.io/metadata.name: "monitoring"
+  egress:
+  - name: "allow-to-my-service"
+    action: "Allow"
+    to:
+    - fqdn:
+        name: "my-service.com"
```
aojea, line 182: is `host.subdomain.my-service.com` a match?

npeps/npep-133-fqdn-egress-selector.md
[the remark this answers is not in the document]
rahulkjoshi: Ok for now, it matches a single level -- we'll leave it that way. We can revisit upon graduation.

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: Whoops, good catch. This was the way I originally wanted to frame it. 
But given the way we defined the network API, I think the discussion in "expected behavior" is more correct.

I'm just removing the section from non-goals.

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: I think the idea we wanted to capture is that we need FQDN Selectors at the namespaced NetworkPolicy layer. This was a suggestion from @danwinship to basically remind ourselves that some sort of FQDN selector will eventually end up in NetPolV2 and we don't want to select any behavior here which can't be implemented in a future implementation.

I'm not sure this user-story does a great job of conveying that... I'll try to think of a better phrasing, but suggestions very much welcome.

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: It is. The reason I split them is we had discussions originally about core vs extended functionality. There was a concern from folks that some NetPol implementations couldn't support wildcards, but could support a direct match.

Splitting the fields gives those implementations a good way to indicate what they support (they can reject policies which set `pattern`). I'm not sure if we still want that approach

@astoycos do you have thoughts? @tssurya as well since I think you did some work with core vs extended definitions

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: I guess we can say they're not supported, but I imagine they're not very useful in this context?

Like if a customer goes from `1.2.3.4` to `my-hostname`, I don't know what we're even allowlisting here. If they need to allow IPs, the network selector lets you specify CIDR blocks.

rahulkjoshi: Oh I think I understand what you mean. I'm adding a section saying that PTR records aren't required to configure the selector. As long as you get an A record, traffic will flow.

If the PTR is missing or even different (e.g. it points to some other domain), the FQDN selector should still work.

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: It's not, the `name` field is supposed to be an exact match. I added some examples to clarify in the API docs.

npeps/npep-133-fqdn-egress-selector.md
npinaeva: If we want to have separate fields for wildcard and non-wildcard matches sounds like a good section for the NPEP. Let's discuss it during the next meeting and write it down for future reference

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,442 @@
...
+    //   - `kubernetes.io` matches only `kubernetes.io`.
+    //     It does not match "www.kubernetes.io", "blog.kubernetes.io",
+    //     "my-kubernetes.io", or "wikipedia.org".
+    //   - `blog.kubernetes.io` matches only "blog.kubernetes.io".
+    //     It does not match "www.kubernetes.io" or "kubernetes.io".
+    //
+    // Support: Extended
+    //
+    // +optional
+    // +kubebuilder:validation:Pattern=`^(([a-zA-z0-9]|[a-zA-Z0-9][-a-zA-Z0-9_]*[a-zA-Z0-9])\.)*([a-zA-z0-9]|[a-zA-Z0-9][-a-zA-Z0-9_]*[a-zA-Z0-9])*\.?$`
```
npinaeva, line 133: I have a suggestions around regex:
`([a-zA-z0-9]|[a-zA-Z0-9][-a-zA-Z0-9_]*[a-zA-Z0-9])` can be replaced with 
`[a-zA-z0-9]([-a-zA-Z0-9_]*[a-zA-Z0-9])?`, do you think it is easier to read or does it miss some valid matches?
I am not sure I understand the last `*` in the expression, I think it can be just removed, otherwise `.` also matches.
This expression matches one-label names, like `com`, do we want to enforce at least 2-label match?

npeps/npep-133-fqdn-egress-selector.md
aojea: it feels weird honestly having both, and this make necessary to treat this as an union, complicating validations, both can not be set at same time

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,442 @@
...
+    //    - `*.*.kubernetes.io` matches subdomains of kubernetes.io at that
+    //      level.
+    //      "sub.subdomain.kubernetes.io" matches, however "kubernetes.io" and
+    //      "blog.kubernetes.io" do not.
+    //
+    // Support: Extended
+    //
+    // +optional
+    // +kubebuilder:validation:Pattern=`^((\*|[a-zA-z0-9]|[a-zA-Z0-9][-a-zA-Z0-9_]*[a-zA-Z0-9])\.)*(\*|[a-zA-z0-9]|[a-zA-Z0-9][-a-zA-Z0-9_]*[a-zA-Z0-9])*\.?$`
+    Pattern string `json:"pattern,omitempty"`
```
fasaxc, line 154: Calico (Enterprise) has DNS policy.  Some thoughts from our impl:

* We only have one field for domain vs pattern so that'd be my preference.
* We only support a single wildcard in the domain but it may match more than one dot-separated part.  
* I think we disallow a redundant trailing `.`.  I think that trailing dot has a subtle meaning in some contexts but since we're not honouring that subtle meaning here I say avoid redundancy.

Our wildcarding works like this:

- *.google.com matches www.google.com and www.ipv6.google.com, but not google.com
- www.*.com matches www.sun.com and www.apple.com, but not www.com
- www.google.* matches www.google.net, www.google.io, and so on
- update.*.mycompany.com matches update.tools.mycompany.com, update.secure.suite.mycompany.com, and so on

We don't allow 

- Multiple wildcards in the same domain, for example: *.*.mycompany.com
- Asterisks that are not the entire component, for example: www.g*.com
- More general wildcards, such as regular expressions

I think those limitations were mainly UI driven but the limit of * only matching a whole segment might allow for more efficient indexing (as opposed to needing to do a regex).

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,442 @@
...
+    //
+    // FQDN is only supported for ALLOW rules. In order to control access, FQDN
+    // ALLOW rules should be used with a lower priority egress deny -- this
+    // allows the admin to maintain an explicit "allowlist" of reachable
+   // domains.
+    //
+    // Support: Extended
+    //
+    // +optional
+    FQDN *FQDNPeerSelector `json:"fqdn,omitempty"`
```
fasaxc, line 114: FQDN feels a bit terse for a k8s API.  Do we lose much by calling this "Domain" or "DomainName" instead?

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,442 @@
...
+      *  Selecting Pods using their [generated DNS
+         record](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pods)
+         (for example `pod-ip-address.my-namespace.pod.cluster.local`). This is
+         analogous to selecting the Pod by its IP address using the Network
+         selector.
+      * Headless Services can be selected using their [generated DNS
+        record](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#services)
+        because the generated DNS records contain a list of all the Pod IPs that
+        back the service.
+   *  ❌ Not Supported:
```
fasaxc, line 306: Just to add another gotcha: we had a user who used DNS load balancing to load balance to a set of external IPs, each in a separate k8s cluster.  They then tried to use DNS policy with _that_; it would work when a pod connected to another cluster but not the local cluster because kube-proxy was short-circuiting the load balancer and handling the external IP in-cluster.  

User thought they were using an external load balancer but that load balancer used external IPs under the covers. Very confusing!

npeps/npep-133-fqdn-egress-selector.md
fasaxc: Calico's policy model only has one field and it feels quite natural in use so that'd be our preference.

npeps/npep-133-fqdn-egress-selector.md
npinaeva: I love these stories :D
so did the problem happen because DNS matching happened after service loadbalancing? IOW, would matching network policy before service resolution solve this problem?

npeps/npep-133-fqdn-egress-selector.md
smuda: Perhaps our user story (which really is a variant of the first user story) would clear anything up?

As a cluster admin, I want to allow certain pods in a namespace to access
an external service specified by a well-known domain name. The service is hosted on a cloud and changes IP-addresses often.

npeps/npep-133-fqdn-egress-selector.md
fasaxc: Perhaps, but then it'd be very hard to interleave normal and DNS policy in anything that resembles iptables.  The DNS policy would happen "first" in order to see the pre-DNAT IPs but what if a higher priority normal policy should have dropped that traffic? BPF dataplanes that do policy and NAT can keep the pre-DNAT information around to match on at the policy stage.

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,442 @@
...
+    // +optional
+    // +kubebuilder:validation:Pattern=`^(([a-zA-z0-9]|[a-zA-Z0-9][-a-zA-Z0-9_]*[a-zA-Z0-9])\.)*([a-zA-z0-9]|[a-zA-Z0-9][-a-zA-Z0-9_]*[a-zA-Z0-9])*\.?$`
+    Name string `json:"names,omitempty"`
+
+    //  Pattern allows matching FQDNs using wildcard specifiers.
+    //
+    // "*" matches 0 or more DNS valid characters (except for "."), and may
+    // occur anywhere in the pattern.
+    //  Examples:
+    //    - `*.kubernetes.io` matches subdomains of kubernetes.io at that level.
```
fasaxc, line 141: I asked my colleagues in our solutions team for examples of matching multiple URL segments, we do have customers that use * to match multiple parts in a prefix.  some examples they were able to share:

* `<team>.<platform>.example.com` matched with `*.example.com`
* `<sessionID>.<session>.<random>.<region>.cache.amazonaws.com` matched with `*.cache.amazonaws.com`

I think the AWS one would be `*.*.*.*.cache.amazonaws.com` in the proposed spec, which feels a bit messy.

npinaeva: very nice examples, thank you for gathering!
I am getting more convinced every day that `*` should match any number of labels, unless someone can point out a security/performance/etc concerns about doing that

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: We discussed this last meeting, and the plan of record is to only support pattern, not name.

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: Updated the regex. Thanks for the suggestions!

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: ack, switched to `domains`.

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: I'm ok with that -- I can see CDNs getting pretty deeply nested so having `*` match multiple labels should be ok.

I don't imagine this should have any performance implications. Security-wise, you sacrifice some explicitness. But on the other hand, when you allow `*.domain.com`, you've already indicated you trust the owner `domain.com`

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: Sounds good to me. For the most part I believe this is compatible with what we're proposing.

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: @smuda I agree, I think this is well captured by the first user story

@astoycos PTAL, I re-worded a little bit, to  hopefully make it easier to understand.

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,428 @@
...
+    // this allows the admin to maintain an explicit "allowlist" of reachable
+    // domains.
+    //
+    // Support: Extended
+    //
+    // <network-policy-api:experimental>
+    // +optional
+    // +listType=set
+    // +kubebuilder:validation:MinItems=1
+    Domains []Domain `json:"domains,omitempty"`
```
astoycos, line 138: In the meeting @npinaeva Proposed `DomainName` here however regardless Antonio suggested we always reference an RFC for decisions like this

fasaxc: https://datatracker.ietf.org/doc/html/rfc1034 S3.1 is the relevant one; 

> A domain is identified by a domain name, and consists of that part of
> the domain name space that is at or below the domain name which
> specifies the domain.

So a _Domain_ is a subtree of domain space whereas a _Domain Name_ is the textual name of one particular node in that tree.  Not sure that gives a clear answer to our naming dilemma though!  `*` patterns represent subtrees, FQDNs represent single nodes in the tree! Perhaps `DomainNamePatterns` conveys exactly what our field represents (but I think `Domains` and `DomainNames` are both defensible/attackable depending on how closely you read the RFC :laughing: )?

The "." suffix is also explained in that section.  A "." at the end of the domain name means it is absolute, whereas no "." means that it is relative to the local domain and the DNS resolver should add suffixes as needed.

npinaeva: I think a domain name with `*` is still considered a domain name. I read RFC as "domain name is a string, domain is an internal structure of the domain name space".
If we read closer to the wildcards definition S4.3.3, it says

> In the previous algorithm, special treatment was given to RRs with owner
names starting with the label "*".

S3.6 defines

> When we talk about a specific RR, we assume it has the following:
owner     -      which is the domain name where the RR is found.

So owner name is the domain name, and owner name may start with `*`, which makes it a valid domain name?
Reading RFCs is fun :D

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,428 @@
...
+1. FQDN policies should not affect the ability of workloads to resolve domains,
+   only their ability to communicate with the IP backing them.
+   *  For example, if a policy allows traffic to `kubernetes.io`, any selected
+      Pods can still resolve `wikipedia.org` or
+      `my-services.default.svc.cluster.local`, but can not send traffic to them
+      unless allowed by a different rule.
+1. Each implementation will provide guidance on which DNS name-server is
+   considered authoritative for resolving domain names. This could be the
+   `kube-dns` Service or potentially some other DNS provider specified in the
+   implementation's configuration.
```
aojea, line 258: does it mean that if implementation X considers kube-dns authoritative but the Pod uses a different nameserver the policy will not apply?

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,428 @@
...
+
+1. A FQDN egress policy does not grant the workload permission to communicate
+   with any in-cluster DNS services (like `kube-dns`). A separate rule needs to
+   be configured to allow traffic to any DNS servers.
+1. FQDN policies should not affect the ability of workloads to resolve domains,
+   only their ability to communicate with the IP backing them.
+   *  For example, if a policy allows traffic to `kubernetes.io`, any selected
+      Pods can still resolve `wikipedia.org` or
+      `my-services.default.svc.cluster.local`, but can not send traffic to them
+      unless allowed by a different rule.
```
aojea, line 254: I kind of infer that this is based on some implementation details, but is hard to get from the text why this will happen ...  "ability of workloads to resolve domains" means to filter DNS traffic or DNS packets? why should the policy filter things are not supposed to do? how is this different than any other traffic?

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,428 @@
...
+1. Each implementation will provide guidance on which DNS name-server is
+   considered authoritative for resolving domain names. This could be the
+   `kube-dns` Service or potentially some other DNS provider specified in the
+   implementation's configuration.
+1. DNS record querying and lifetimes:
+   *  Pods are expected to make a DNS query for a domain before sending traffic
+      to it. If the Pod fails to send a DNS request and instead just sends
+      traffic to the IP (either because of caching or a static config), traffic
+      is not guaranteed to flow.
```
aojea, line 263: is this implementable? this means the implementation has to cache all possible IPs of all the possible domains, no?

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,428 @@
...
+   *  Pods should respect the TTL of DNS records they receive. Trying to
+      establish new connection using DNS records that are expired is not
+      guaranteed to work.
+   *  When the TTL for a DNS record expires, the implementor should stop
```
aojea, line 267: what TTL records is referring here? the TTL records of the DNS response that the Pod has received?
does it imply always the implementation to snoop?

aojea: ok, I see the table below, good work with that table, really intuitive

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,428 @@
...
+        www.kubernetes.io  --  CNAME to kubernetes.io
+        
+        # REQUEST 2
+
+        kubernetes.io      --  A     to 1.2.3.4
+        ```
+        The implementer can still deny traffic to `1.2.3.4` because no single
+        response contained the full chain required to resolve the domain.
+
+## Alternatives
```
aojea, line 333: Other alternative is implement DNS filtering,  as cisco umbrella https://docs.umbrella.com/deployment-umbrella/docs/point-your-dns-to-cisco or  dnsfilter do ...

npeps/npep-133-fqdn-egress-selector.md
illrill: In https://github.com/cilium/cilium/issues/22081, I've been discussing this topic with maintainers of Cilium (wich regards to `CiliumNetworkPolicy`) . Though still not decided or implemented, the dialog has been leaning towards:

  - `**` matches any character (including `.`). Allows wildcarding of multiple levels of subdomains.
  - `*` matches any character except `.`. Allows wildcarding of a single subdomain.

Perhaps this can be useful, if not just for some additional perspectives.

npeps/npep-133-fqdn-egress-selector.md
fasaxc: They do say that the owner is a domain(!) but they also say 
```
If any match, copy them into the answer
section, but set the owner of the RR to be QNAME, and
not the node with the "*" label.
```
I.e. don't send `*` on the wire and 
```
The owner name of the wildcard RRs is of
the form "*.<anydomain>", where <anydomain> is any domain name.
<anydomain> should not contain other * labels
```
Which implicitly says that a wildcard is not a domain so, it's a bit contradictory!

I think the wildcard comes from implementation concerns, they're really talking about the configuration for a DN server (and probably a particular one at that) and saying "you'll need wildcards in your implementation to handle mail and they should work like this to avoid these problems".

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: I'm trying to clarify that this spec is explicitly **not** doing DNS filtering. So even though the policy says "allow traffic to kubernetes.io", you can still resolve any domains you like.

I can add a clarification explicitly saying that FQDN policies will not perform DNS Filtering.

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: As long as the two nameservers return the same IPs for domains it's not a problem. But if kube-dns says the IP is `1.2.3.4`, the implementation only allows `1.2.3.4`. If the pod's nameserver then says `5.6.7.8`, then traffic will be denied.

npeps/npep-133-fqdn-egress-selector.md
```diff
@@ -0,0 +1,428 @@
...
+//      It does not match "www.kubernetes.io", "blog.kubernetes.io",
+//      "my-kubernetes.io", or "wikipedia.org".
+//    - `blog.kubernetes.io` matches only "blog.kubernetes.io".
+//      It does not match "www.kubernetes.io" or "kubernetes.io".
+//    - `*.kubernetes.io` matches subdomains of kubernetes.io.
+//      "www.kubernetes.io", "blog.kubernetes.io", and
+//      "latest.blog.kubernetes.io" match, however "kubernetes.io", and
+//      "wikipedia.org" do not.
+//
+// +kubebuilder:validation:Pattern=`^(\*\.)?([a-zA-z0-9]([-a-zA-Z0-9_]*[a-zA-Z0-9])?\.)*[a-zA-z0-9]([-a-zA-Z0-9_]*[a-zA-Z0-9]))?\.?$`
```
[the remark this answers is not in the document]
rahulkjoshi: Thanks for the great suggestion. Done!

npeps/npep-133-fqdn-egress-selector.md
rahulkjoshi: Yes we've pretty much agreed that the only way to correctly implement this is via a proxy that listens to DNS responses.

Issue #220, opened 2024-04-18
[Policy Assistant] user experience for specifying pods/traffic


huntergregory:
Parent issue: #150 

## TL;DR

Write go code to get a Deployment/DaemonSet from a cluster and create a corresponding `TrafficPeer` (see struct referenced below).

## Overview

Our KubeCon demo ([video](https://youtu.be/riSv0g-TNtI?si=6gxDON4jK2xf4n-e&t=1449), [slides](https://static.sched.com/hosted_files/kccnceu2024/bc/KubeConEU24_NetworkPolicy_API.pptx.pdf), [code](https://github.com/huntergregory/network-policy-api/tree/kubecon/cmd/policy-assistant/kubecon)) used hard-coded Pods and traffic (Pod a to Pod b on port 80 and 81).
![image](https://github.com/kubernetes-sigs/network-policy-api/assets/42728408/688b4f52-82b4-44b8-8490-5e54e6739093)

Of course, a user should be able to specify their own Pods/traffic. Ideally, a user can:
- Quickly/easily specify traffic of interest.
- Reference Pods in a Kubernetes-native way.

## Ideas

A helpful starting point is Matt's [Traffic struct](https://github.com/kubernetes-sigs/network-policy-api/blob/669dfbc72245a2c7c6f3a3dd5c6188b7af9f8014/cmd/policy-assistant/pkg/matcher/traffic.go#L13). Here's a snippet of this [example json](https://github.com/kubernetes-sigs/network-policy-api/blob/main/cmd/policy-assistant/examples/traffic.json):
```json
{
  "Source": {
    "IP": "8.8.8.8"
  },
  "Destination": {
    "Internal": {
      "PodLabels": {"pod": "a"},
      "NamespaceLabels": {"ns":  "y"},
      "Namespace": "y"
      // we will also need NodeLabels for AdminNetworkPolicy's node selector
    },
    "IP": "192.168.1.10"
  },
  "Protocol": "TCP",
  "ResolvedPort": 80,
  "ResolvedPortName": "serve-80-tcp"
}
```

**_It would be nice if a user could instead reference a Pod/Deployment/DaemonSet_**, and then Policy Assistant queries someone's cluster to fill in:
- pod labels
- namespace labels
- node labels
- IP (or IPs of a Deployment, for instance)

We could start by building go code to convert a Deployment or DaemonSet to a `TrafficPeer` for a user's Kubernetes cluster.

[...]

Issue #168, opened 2023-11-01
[Policy Assistant] Predict Verdict and Flow of Decisions made by Policies


huntergregory:
Idea for #150.

Dependency: #152

Predict the effect of Policies on traffic, and visualize the relevant rules (from ANP, NetPol v1, and/or BANP) that produced the effect.

Prototype visualization:
![image](https://github.com/kubernetes-sigs/network-policy-api/assets/42728408/86511727-273a-4543-89c3-cd4fab797363)

gabrielggg:
Hi @huntergregory can this be sumarized as: "Repurposing cyclonus analyze --mode=query-traffic" ?

huntergregory:
Hey @gabrielggg, good call out. We're still in a stage of design iteration for this, and `query-traffic` mode may be a helpful reference for design or implementation. For reference, `query-traffic` mode looked like this for NetworkPolicyV1 ([full output](https://github.com/mattfenwick/cyclonus/blob/master/examples/cyclonus-output/analyze-query-traffic.txt)):
<img width="425" alt="image" src="https://github.com/kubernetes-sigs/network-policy-api/assets/42728408/5961b43f-56e4-400b-b733-02ebbdf554f7">

On the other hand, this is what we came up with for the KubeCon demo ([video](https://youtu.be/riSv0g-TNtI?si=VJLV5b9MJtsIZHJd&t=1234) and [code](https://github.com/huntergregory/network-policy-api/tree/kubecon/cmd/policy-assistant/kubecon)):
<img width="809" alt="image" src="https://github.com/kubernetes-sigs/network-policy-api/assets/42728408/c209ead6-66d2-471e-8e55-08381a35a9c5">

gabrielggg:
hey @huntergregory awesome, first of all, congratulations on the great presentation on kubecon Paris, i just saw the video. At the end of the video, on the summary, you put a slide with the topics coming next on the project, and i saw that you highlighted this issue #168 , but seeing the demo and your previous screenshots i see that the walkthrough you proposed on this issue is almost done, so i just wanted to know how are you going to proceed with this issue, so that i can help.

huntergregory:
Thanks @gabrielggg for the congrats and the interest, and sorry for the delay.

Just created #220, which is the main thing lacking right now. We could definitely use some help in this area! Feel free to assign yourself to that issue.

[...]

Issue #221, opened 2024-04-19
[Policy Assistant] calculate all allowed connections in a cluster


huntergregory:
Parent issue: #150 

## Goal

For all the Deployments and DaemonSets in a cluster, calculate the set of allowed connections given a set of policies.

## Current Functionality
Check whether traffic is allowed/denied **_for the specified source, destination, and port/protocol_**.

![image](https://github.com/kubernetes-sigs/network-policy-api/assets/42728408/863f1238-4b66-4a6d-922e-19efad4ce842)

## Proposed New Feature

Produce JSON of all allowed connections and the effective policy rules causing this.

Here is one idea for the format (let me put this in a PR):
```json
{
    "Connections": [
        {
            "From": "10.0.0.0/28",
            "To": "ns-dev/daemonset/backend",
            "AllowedTraffic": {
                "TCPPorts": "all"
            },
            "EffectivePolicyRules": {
                "Ingress": {
                    "ANPPassed": [
                        {
                            "PolicyKey": "default/anp1",
                            "Priority": 1,
                            "RuleNumber": 9,
                            "Rule": "dev-pass-on-external-tcp",
                            "TCPPorts": "all"
                        }
                    ],
                    "ANPDenied": [
                        {
                            "PolicyKey": "default/anp1",
                            "Priority": 1,
                            "RuleNumber": 10,
                            "Rule": "deny-external",
                            "UDPPorts": "all",
                            "OtherPorts": "all"
                        }
                    ],
                    "NPv1Allowed": [
                        {
                            "PolicyKey": "ns-dev/allow-all",
                            "TCPPorts": "all"
                        }
                    ]
                }
            }
        },
        {
            "From": "ns-dev/deployment/frontend",
            "To": "ns-dev/daemonset/backend",
            "AllowedTraffic": {
                "TCPPorts": "80",
                "UDPPorts": "53,80"
            },
            "EffectivePolicyRules": {
                "Ingress": {
                    "ANPAllowed": [
                        {
                            "PolicyKey": "default/anp2",
                            "Priority": 2,
                            "RuleNumber": 1,
                            "Rule": "allow-all-udp",
                            "UDPPorts": "all"
                        }
                    ],
                    "ANPPassed": [
                        {
                            "PolicyKey": "default/anp1",
                            "Priority": 1,
                            "RuleNumber": 1,
                            "Rule": "pass-tcp-80",
                            "TCPPorts": "80"
                        },
                        {
[...]

Issue #170, opened 2023-11-06
[Policy Assistant] use internal vendoring instead of vendoring from Cyclonus


huntergregory:
Related to #150.

See https://github.com/kubernetes-sigs/network-policy-api/pull/159#discussion_r1374029219

PR #227, opened 2024-04-26
[Policy Assistant] Add support for k8s native workload traffic

gabrielggg:
Hi @huntergregory as we discussed on #168 , this is to solve #220 , i've implemented this so that a user can use both types of traffic (native, non native) in a same traffic.json file (please refer to https://github.com/gabrielggg/network-policy-api/blob/main/cmd/policy-assistant/examples/traffic.json file to see some examples). Now the trafficPeer is supporting both. I hope you like this and give some feedback.

For example, for a traffic input like this one :

{
    "Source": {
      "Internal": {
        "Workload": {"daemonset": "fluentd-elasticsearch"},
        "Namespace": "kube-system"
      }
    },
    "Destination": {
      "Internal": {
        "Workload": {"deployment": "nginx-deployment2"},
        "Namespace": "default"
      }
    },
    "Protocol": "TCP",
    "ResolvedPort": 80,
    "ResolvedPortName": "serve-80-tcp"
  },{
    "Source": {
      "Internal": {
        "Workload": {"daemonset": "fluentd-elasticsearch"},
        "Namespace": "kube-system"
      }
    },
    "Destination": {
      "Internal": {
        "PodLabels": {"pod": "b"},
        "NamespaceLabels": {"ns": "y"},
        "Namespace": "y"
      },
      "IP": "192.168.1.100"
    },
    "Protocol": "TCP",
    "ResolvedPort": 80,
    "ResolvedPortName": "serve-80-tcp"
  }

You get this output (considering that both the deployment and the daemonset have 2 replicas):

image
image
image

Please check this out!

### DIFF
--- a/cmd/policy-assistant/pkg/matcher/traffic.go
+++ b/cmd/policy-assistant/pkg/matcher/traffic.go
@@ -5,7 +5,10 @@ import (
 	"strings"
 
 	"github.com/mattfenwick/collections/pkg/slice"
+	"github.com/mattfenwick/cyclonus/pkg/kube"
+	"github.com/mattfenwick/cyclonus/pkg/utils"
 	"github.com/olekukonko/tablewriter"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/exp/maps"
 	v1 "k8s.io/api/core/v1"
 )
@@ -57,7 +60,8 @@ func labelsToString(labels map[string]string) string {
 
 type TrafficPeer struct {
 	Internal *InternalPeer
-	IP       string
+	// IP external to cluster
+	IP string
 }
 
 func (p *TrafficPeer) Namespace() string {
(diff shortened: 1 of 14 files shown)


huntergregory (changes requested):
Hi @gabrielggg, thanks for the PR. You are speedy! ⚡😅 Sorry that I was not as quick to respond.

Taking a step back, I think it'd be easiest to finalize the data structures (and how to populate them) before starting the code that will use the data structures.

These are the parts of #220 that I'd recommend we focus on first:
> TL;DR
Write go code to get a Deployment/DaemonSet from a cluster and create a corresponding TrafficPeer (see struct referenced below).
...
It would be nice if a user could instead reference a Pod/Deployment/DaemonSet, and then Policy Assistant queries someone's cluster to fill in:
\- pod labels
\- namespace labels
\- node labels
\- IP (or IPs of a Deployment, for instance)
We could start by building go code to convert a Deployment or DaemonSet to a TrafficPeer for a user's Kubernetes cluster.

### Data Format
I think your JSON design is nearly there. The main caveat with it is that it will be hard to track different IPs for the Pods of a workload.

So specifically, I would probably convert these structs
```go
type TrafficPeer struct {
	Internal *InternalPeer
	IP          string
}

type InternalPeer struct {
	PodLabels       map[string]string
	NamespaceLabels map[string]string
	Namespace       string
	NodeLabels      map[string]string
	Node            string
}
```
to
```go
type TrafficPeer struct {
	Internal *InternalPeer
       // keep this field for backwards-compatibility or for IPs without internalPeer
	IP          string
       // use this for pod IPs
       *Workload
}

type InternalPeer struct {
	PodLabels       map[string]string
	NamespaceLabels map[string]string
	Namespace       string
       // I believe I added these node pieces. We can remove
}

type Workload struct {
      // format: namespace/kind/name
      
       pods []PodNetworking
}

type PodNetworking struct {
       IP string
      // don't worry about populating below fields right now
       IsHostNetworking bool
       NodeLabels []string
}
```

Feel free to suggest tweaks as well. For instance, I think the placement of the workload name field is somewhat arbitrary and could always be refactored to make using the structs easier.

### Populating the Data from Cluster
After we finalize the structs, would you be able to make those modifications? Could you please also see where `TrafficPeer.IP` is referenced? We can worry about what to do where it is referenced later.

Then, would you be ok writing go code to translate a user input (say "ns-dev/deployment/frontend") into your modified `TrafficPeer` struct? I would recommend we don't modify any code for the `analyze` command in this first PR.

huntergregory:
Hey @gabrielggg I accidentally published my review before I finished writing. When you can, please take a look at the edited text above and let me know how this sounds

gabrielggg:
Hey @huntergregory , thanks a lot for the review and the guidance, i just have one question regarding the review; Is there a particular reason for putting the Workload struct outside of the InternalPeer struct? I imagine the workloads being always categorized as internal traffic. 

Maybe something like the this makes sense?

`type TrafficPeer struct {
	Internal *InternalPeer
       // keep this field for backwards-compatibility or for IPs without internalPeer
	IP          string
}

type InternalPeer struct {
	PodLabels       map[string]string
	NamespaceLabels map[string]string
	Namespace       string
        // I believe I added these node pieces. We can remove
        // use this for pod IPs
        *Workload
}

type Workload struct {
      // format: namespace/kind/name
       fullName string
       pods []PodNetworking
}

type PodNetworking struct {
       IP string
      // don't worry about populating below fields right now
       IsHostNetworking bool
       NodeLabels []string
}`

gabrielggg:
```go
test
```

huntergregory:
Hey @gabrielggg, taking a second look I’d approve your original idea, and my only suggestion is adding the pods field to InternalPeer. Let’s keep everything internal (as in internal to cluster). So like:

```go
type TrafficPeer struct {
	Internal *InternalPeer
        // IP external to cluster
	IP          string
}

// Internal to cluster
type InternalPeer struct {
        // optional: if set, will override remaining values with information from cluster
        workload string

	PodLabels       map[string]string
	NamespaceLabels map[string]string
	Namespace       string
        // optional
        Pods      []*PodNetworking
}
```

Still would prefer to save the traffic command logic for a follow-up PR if you don’t mind? So this PR would just modify data structures and define a function that populates InternalPeer based on workload name.

gabrielggg:
Hey @huntergregory , thanks a lot for your review and guidance on this PR, it helped a lot!

Well, i implemented the new structs and created a func to Translate the TrafficPeer, as you suggested , i'm not going to touch the analyze command on this PR, i think it makes sense to treat that separately.

After some testing of the translate function i'm getting the following object.

For this traffic example (nginx deployment workload has 2 replicas):

```go
[
    {
      "Source": {
        "Internal": {
            "Workload": "default/deployment/nginx-deployment2"
          }
        
      },
      "Destination": {
        "IP": "8.8.8.8"
      },
      "Protocol": "TCP",
      "ResolvedPort": 80,
      "ResolvedPortName": "serve-80-tcp"
    }   
]
``` 
Then after doing something like this:

```go
b, err := json.Marshal(traffic.Source.Translate())
if err != nil {
fmt.Println(err)
return
}
fmt.Println(string(b))	
``` 

I'm getting this translated TrafficPeer object returned:

```go
{
  "Internal": {
    "Workload": "default/deployment/nginx-deployment2",
    "PodLabels": {
      "app": "nginx",
      "app2": "tmp",
      "pod-template-hash": "5c496748cb"
    },
    "NamespaceLabels": {
      "kubernetes.io/metadata.name": "default",
      "labelname2": "value2"
    },
    "Namespace": "default",
    "Pods": [
      {
        "IP": "10.224.0.41",
        "IsHostNetworking": false,
        "NodeLabels": null
      },
      {
        "IP": "10.224.0.234",
        "IsHostNetworking": false,
        "NodeLabels": null
      }
    ]
  },
  "IP": ""
}
``` 

Please let me know what do you think about this!

huntergregory (changes requested):
Thanks @gabrielggg this is awesome. Would you mind supporting a basic Pod type too? Other than that, just a couple minor comments.

Also do you mind putting the yamls you're using for testing in the policy assistant's `test/` directory? And putting the input/output JSONs in separate files too? Ideally, we could use these to create an e2e test someday.

huntergregory:
By the way, the next capability I had in mind for Policy Assistant is a continuation of this work. For #221, I think we should get all Deployments and DaemonSets in the cluster, then convert them to TrafficPeers. Maybe later we could support more than Deployment and DaemonSet.

Would you be able to help with this small task too since it is similar? Maybe we can reuse code from these other functions? I would be ok with doing that work in this PR too

gabrielggg:
Hey @huntergregory , thanks a lot for the review and comments, i think i solved all the comments you made on the code (including support for pods and replicaset), please check it out and tell me what you think; I will also upload the input/output json files from my tests and the yamls from the workloads used for testing as you suggested.

Regarding your last comment, i will be working on that on this PR next.

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -71,10 +75,80 @@ func (p *TrafficPeer) IsExternal() bool {
...
 }
 
+func (p *TrafficPeer) Translate() TrafficPeer {
+	var podsNetworking []*PodNetworking
+	var podLabels map[string]string
+	var namespaceLabels map[string]string
+	var workloadOwner string
+	workloadOwnerExists := false
+	workloadMetadata := strings.Split(strings.ToLower(p.Internal.Workload), "/")
+	if (workloadMetadata[0] == "" || workloadMetadata[1] == "" || workloadMetadata[2] == "") || len(workloadMetadata) != 3 || (workloadMetadata[1] != "daemonset" && workloadMetadata[1] != "statefulset" && workloadMetadata[1] != "replicaset" && workloadMetadata[1] != "deployment" && workloadMetadata[1] != "pod") {
```
huntergregory, line 85: nit: think this code could crash if `p.Internal.Workload = "abc"` for instance

Could move `len(workloadMetadata) != 3` first to make the boolean short-circuit

huntergregory (commented):
Thanks this looks great!

cmd/policy-assistant/pkg/matcher/traffic.go
gabrielggg: Thanks, yes, true. I fixed it.

huntergregory:
Hey @gabrielggg, were you still thinking to convert all Deployments/DaemonSets to TrafficPeers in this PR?

gabrielggg:
Hi, @huntergregory , yes i'm working on it.

gabrielggg:
Hey @huntergregory , sorry for the delay, it's done, i created 2 functions (1 for mapping deployments and 1 for mapping daemonsets; Maybe we can support othe workload types also), both are using the previously create Translate func and return a slice containing all the generated TrafficPeers, please check it out!

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -71,10 +75,245 @@ func (p *TrafficPeer) IsExternal() bool {
...
+		}
+		for _, deployment := range kubeDeployments {
+			TmpInternalPeer := InternalPeer{
+				Workload: namespace.Name + "/deployment/" + deployment.Name,
+			}
+			TmpPeer := TrafficPeer{
+				Internal: &TmpInternalPeer,
+			}
+			TmpPeerTranslated := TmpPeer.Translate()
+			if TmpPeerTranslated.Internal.Workload != "" {
```
huntergregory, line 164: related to `Translate()` returning an error

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -71,10 +75,245 @@ func (p *TrafficPeer) IsExternal() bool {
...
+		logrus.Fatalf("unable to read namespaces from kube: %+v", err)
+	}
+
+	for _, namespace := range kubeNamespaces.Items {
+		kubeDeployments, err := kubeClient.GetDeploymentsInNamespace(namespace.Name)
+		if err != nil {
+			logrus.Fatalf("unable to read deployments from kube, ns '%s': %+v", namespace.Name, err)
+		}
+		for _, deployment := range kubeDeployments {
+			TmpInternalPeer := InternalPeer{
```
huntergregory, line 157: nit: [conventional in go](https://medium.com/@singhalok641/mastering-exported-and-unexported-names-in-go-a-key-to-effective-encapsulation-6d4b28bd61a6) to use lowercase for variables in functions

huntergregory (commented):
No worries @gabrielggg. This is nice how we can reuse the `Translate()` function! Thanks for taking initiative to do the other workload types too. If we're doing all workloads, I was envisioning a function that gets all Pods in the cluster and creates only one TrafficPeer per Pod. Would you mind modifying your code to support this? Maybe handle all Deployments and DaemonSets first, then remaining ReplicaSets and StatefulSets, and finally remaining Pods.

Could you add results of testing this in the `test/` directory? And if you don't mind, let's also add test yamls for a basic Pod, StatefulSet, and ReplicaSet (yamls which are not associated with a Deployment or DaemonSet).

This is awesome to see. Everything's coming together nicely 😁

> I've created a separated func for each workload type. But maybe it will be better to put everything on 1 func to reuse the code. Please tell me what you think about it.

Looks ok to me but up to you what seems easier to write/read/maintain 🙂

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -71,10 +75,245 @@ func (p *TrafficPeer) IsExternal() bool {
...
+			podsNetworking = append(podsNetworking, &podNetworking)
+			workloadOwnerExists = true
+
+		}
+	}
+
+	if !workloadOwnerExists {
+		logrus.Infof("workload not found on the cluster")
+		internalPeer = InternalPeer{
+			Workload: "",
```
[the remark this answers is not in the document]
gabrielggg: hey @huntergregory , thanks for the review, the problem here is that doing a fatal exits my code when a workload is scaled down to 0. for example if you have a deployment with 0 replicas it crashes there when you call the function to map all the deployments on the cluster to trafficpeers. Have you thought about that scenario.

gabrielggg:
Hey @huntergregory, thanks a lot for the review, i'm working on the observations but i have one doubt regarding this comment  :

>I was envisioning a function that gets all Pods in the cluster and creates only one TrafficPeer per Pod. Would you mind modifying your code to support this? Maybe handle all Deployments and DaemonSets first, then remaining ReplicaSets and StatefulSets, and finally remaining Pods.
> 

 The PodsToTrafficPeers() function does exactly that right now, it create a TrafficPeer per pod on the cluster. Maybe i'm not understanding what do you mean, if so, can you please elaborate a little bit. Thanks a los in advance.

huntergregory:
> > I was envisioning a function that gets all Pods in the cluster and creates only one TrafficPeer per Pod. Would you mind modifying your code to support this? Maybe handle all Deployments and DaemonSets first, then remaining ReplicaSets and StatefulSets, and finally remaining Pods.
> 
> The PodsToTrafficPeers() function does exactly that right now, it create a TrafficPeer per pod on the cluster. Maybe i'm not understanding what do you mean, if so, can you please elaborate a little bit. Thanks a los in advance.

Hey sorry @gabrielggg, I expressed this incorrectly. I mean for each Pod to be _associated_ with exactly one TrafficPeer. For example, all Pods in a Deployment should be associated with the same TrafficPeer (with the Deployment workload type). Those same Pods should not be associated with a TrafficPeer with the ReplicaSet or Pod workload type. So there should be less TrafficPeers than Pods. The function logic would be something like:
1. Create one TrafficPeer for each Deployment, and consider all Pods in the Deployment as "handled".
2. Similar for DaemonSets.
3. Get all ReplicaSets. Ignore the ReplicaSet if it's associated with a Deployment (equally, ignore Pods that have been "handled" already).
4. Similar for StatefulSets.
5. For remaining Pods, only create a TrafficPeer if a Pod has not been "handled".

cmd/policy-assistant/pkg/matcher/traffic.go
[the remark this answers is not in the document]
huntergregory: My suggestion is nitpicky, and I'm noticing it requires a good amount of change, so honestly feel free to leave everything as is 🙂. I was more so suggesting that we could change the function signature to:
```go
func (p *TrafficPeer) Translate() (*TrafficPeer, error) {
```

and instead of using `logrus.Fatal()`, we could start returning errors to be handled as needed:
```go
return nil, fmt.Errorf("failed to get workload: %w", err)
```

But this might be a lot of work for little reward: if you replace `logrus.Fatal()` in this function, you might as well replace it in all functions. So feel free to leave as is 🙂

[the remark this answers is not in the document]
huntergregory: > for example if you have a deployment with 0 replicas

Good call out. Do you mind calling this out somehow in the code too?

gabrielggg:
thanks for the clarification @huntergregory , I think that makes sense. I will be working on it.

gabrielggg:
hey @huntergregory , i managed to get each Pod to be associated with exactly one TrafficPeer. Please check that out, also added all the test files used for testing including workload yamls and json outputs of each function, please check it out to and tell me what you think.

cmd/policy-assistant/pkg/matcher/traffic.go
gabrielggg: Here i think we are not going to return an error based on the conversation we had.

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -5,7 +5,10 @@ import (
 	"strings"
 
 	"github.com/mattfenwick/collections/pkg/slice"
+	"github.com/mattfenwick/cyclonus/pkg/kube"
```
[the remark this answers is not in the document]
gabrielggg: I think this is related to this issue https://github.com/kubernetes-sigs/network-policy-api/issues/170

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -250,16 +253,22 @@ func ReplicaSetsToTrafficPeers() []TrafficPeer {
...
-				Workload: namespace.Name + "/replicaset/" + replicaSet.Name,
-			}
-			TmpPeer := TrafficPeer{
-				Internal: &TmpInternalPeer,
-			}
-			TmpPeerTranslated := TmpPeer.Translate()
-			if TmpPeerTranslated.Internal.Workload != "" {
-				replicaSetPeers = append(replicaSetPeers, TmpPeerTranslated)
+			if replicaSet.OwnerReferences != nil {
+				logrus.Infof("replicaset already handled")
```
huntergregory, line 259: nit: this could be a noisy log
```suggestion
				continue
```

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -71,10 +75,258 @@ func (p *TrafficPeer) IsExternal() bool {
...
+				statefulSetPeers = append(statefulSetPeers, tmpPeerTranslated)
+			}
+		}
+
+	}
+
+	return statefulSetPeers
+}
+
+func ReplicaSetsToTrafficPeers() []TrafficPeer {
```
huntergregory, line 242: nit: could you add comments describing the functions? At least for this one and `PodsToTrafficPeers()`. Including the fact that it ignores Pods associated with Deployments, etc.

cmd/policy-assistant/pkg/matcher/traffic.go
huntergregory: same thing in `PodsToTrafficPeers()`

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -71,10 +75,258 @@ func (p *TrafficPeer) IsExternal() bool {
...
+		if workloadMetadata[1] == "deployment" && pod.OwnerReferences != nil && pod.OwnerReferences[0].Kind == "ReplicaSet" {
+			kubeReplicaSets, err := kubeClient.GetReplicaSet(workloadMetadata[0], pod.OwnerReferences[0].Name)
+			if err != nil {
+				logrus.Fatalf("unable to read Replicaset from kube, rs '%s': %+v", pod.OwnerReferences[0].Name, err)
+			}
+			if kubeReplicaSets.OwnerReferences != nil {
+				workloadOwner = kubeReplicaSets.OwnerReferences[0].Name
+			}
+
+		} else if (workloadMetadata[1] == "daemonset" || workloadMetadata[1] == "statefulset" || workloadMetadata[1] == "replicaset") && pod.OwnerReferences != nil {
```
huntergregory, line 107: realizing there's an edge case when there is a DaemonSet, StatefulSet, or ReplicaSet sharing the same name?

Check `pod.OwnerReferences[0].Kind` here as well?

huntergregory (commented):
Nice! Love that it was a simple solution 🙂. Thanks for adding all of the yamls/outputs too.

I just noticed a tiny edge case and had two last nitpicks.

gabrielggg:
hey @huntergregory , thanks for the review. I fixed the edge and removed the log. Please check it out.

cmd/policy-assistant/pkg/matcher/traffic.go
gabrielggg: comments added

cmd/policy-assistant/pkg/matcher/traffic.go
gabrielggg: done

huntergregory (approved):
Thank you @gabrielggg for the HUGE help and persistence. This will be key infra for Policy Assistant going forward.

/lgtm 🚀

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -71,10 +75,268 @@ func (p *TrafficPeer) IsExternal() bool {
...
+	workloadOwnerExists := false
+	workloadMetadata := strings.Split(strings.ToLower(p.Internal.Workload), "/")
+	if len(workloadMetadata) != 3 || (workloadMetadata[0] == "" || workloadMetadata[1] == "" || workloadMetadata[2] == "") || (workloadMetadata[1] != "daemonset" && workloadMetadata[1] != "statefulset" && workloadMetadata[1] != "replicaset" && workloadMetadata[1] != "deployment" && workloadMetadata[1] != "pod") {
+		logrus.Fatalf("Bad Workload structure: Types supported are pod, replicaset, deployment, daemonset, statefulset, and 3 fields are required with this structure, <namespace>/<workloadType>/<workloadName>")
+	}
+	kubeClient, err := kube.NewKubernetesForContext("")
+	utils.DoOrDie(err)
+	ns, err := kubeClient.GetNamespace(workloadMetadata[0])
+	utils.DoOrDie(err)
+	kubePods, err := kube.GetPodsInNamespaces(kubeClient, []string{workloadMetadata[0]})
```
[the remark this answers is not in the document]
gabrielggg: Ok @huntergregory , no problem, i will create the issue and also squash the commits on this PR.

gabrielggg:
closes #220

PR #229, opened 2024-05-14
Update the text about duplicate priority values

### LINKED ISSUE #216: having two ANPs with the same priority must be well-defined
It's unacceptable to say that the behavior when you have two ANPs with the same priority is completely undefined.

We can say that it's undefined which order they are applied in, but it must be the case that it is guaranteed that both are applied, and that two policies with `priority: N` are both applied in between policies with `priority: N+1` and policies with `priority: N-1`.

danwinship:
Clarify the limits of the undefined behavior when two policies have the same priority.

Fixes #216

/cc @astoycos @tssurya @npinaeva @fasaxc @nathanjsweet @aojea

### DIFF
--- a/apis/v1alpha1/adminnetworkpolicy_types.go
+++ b/apis/v1alpha1/adminnetworkpolicy_types.go
@@ -56,11 +56,14 @@ type AdminNetworkPolicyStatus struct {
 
 // AdminNetworkPolicySpec defines the desired state of AdminNetworkPolicy.
 type AdminNetworkPolicySpec struct {
-	// Priority is a value from 0 to 1000. Rules with lower priority values have
-	// higher precedence, and are checked before rules with higher priority values.
+	// Priority is a value from 0 to 1000. Policies with lower priority values have
+	// higher precedence, and are checked before policies with higher priority values.
 	// All AdminNetworkPolicy rules have higher precedence than NetworkPolicy or
 	// BaselineAdminNetworkPolicy rules
-	// The behavior is undefined if two ANP objects have same priority.
+	// Every AdminNetworkPolicy should have a unique priority value; if two (or more)
+	// policies with the same priority could both match a connection, then the
+	// implementation can apply any of the matching policies to the connection, and
+	// there is no way for the user to reliably determine which one it will choose.
 	//
 	// Support: Core
 	//
--- a/config/crd/experimental/policy.networking.k8s.io_adminnetworkpolicies.yaml
+++ b/config/crd/experimental/policy.networking.k8s.io_adminnetworkpolicies.yaml
@@ -817,11 +817,14 @@ spec:
                 type: array
               priority:
                 description: |-
-                  Priority is a value from 0 to 1000. Rules with lower priority values have
-                  higher precedence, and are checked before rules with higher priority values.
+                  Priority is a value from 0 to 1000. Policies with lower priority values have
+                  higher precedence, and are checked before policies with higher priority values.
                   All AdminNetworkPolicy rules have higher precedence than NetworkPolicy or
                   BaselineAdminNetworkPolicy rules
-                  The behavior is undefined if two ANP objects have same priority.
+                  Every AdminNetworkPolicy should have a unique priority value; if two (or more)
+                  policies with the same priority could both match a connection, then the
+                  implementation can apply any of the matching policies to the connection, and
+                  there is no way for the user to reliably determine which one it will choose.
 
 
                   Support: Core
--- a/config/crd/standard/policy.networking.k8s.io_adminnetworkpolicies.yaml
+++ b/config/crd/standard/policy.networking.k8s.io_adminnetworkpolicies.yaml
@@ -701,11 +701,14 @@ spec:
                 type: array
               priority:
                 description: |-
-                  Priority is a value from 0 to 1000. Rules with lower priority values have
-                  higher precedence, and are checked before rules with higher priority values.
+                  Priority is a value from 0 to 1000. Policies with lower priority values have
+                  higher precedence, and are checked before policies with higher priority values.
                   All AdminNetworkPolicy rules have higher precedence than NetworkPolicy or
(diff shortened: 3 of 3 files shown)


apis/v1alpha1/adminnetworkpolicy_types.go
```diff
@@ -60,7 +60,9 @@ type AdminNetworkPolicySpec struct {
 	// higher precedence, and are checked before rules with higher priority values.
 	// All AdminNetworkPolicy rules have higher precedence than NetworkPolicy or
 	// BaselineAdminNetworkPolicy rules
-	// The behavior is undefined if two ANP objects have same priority.
+	// Every AdminNetworkPolicy should have a unique priority value. If multiple
+	// policies have the same priority, and match the same connection, then it is
+	// implementation-defined which policy's rules get applied to it.
```
npinaeva, line 65: maybe replace "implementation-defined" with "the policies will be applied in some order"? implementation-defined feels like implementations must tell you an order, but I think the point is to not rely on any specific order here

danwinship: I said "implementation-defined" because I wanted to make it clear that different implementations might do different things. I also didn't want to say anything that could be read as implying that the result will be _predictable_...

Updated... what do you think of the new text?

apis/v1alpha1/adminnetworkpolicy_types.go
```diff
@@ -56,11 +56,14 @@ type AdminNetworkPolicyStatus struct {
...
-	// higher precedence, and are checked before rules with higher priority values.
+	// Priority is a value from 0 to 1000. Policies with lower priority values have
+	// higher precedence, and are checked before policies with higher priority values.
 	// All AdminNetworkPolicy rules have higher precedence than NetworkPolicy or
 	// BaselineAdminNetworkPolicy rules
-	// The behavior is undefined if two ANP objects have same priority.
+	// Every AdminNetworkPolicy should have a unique priority value; if two (or more)
+	// policies with the same priority could both match a connection, then there is no
+	// way for the user to reliably determine which of the policies will actually be
+	// applied to it.
```
tssurya, line 66: looks nicer now than saying implementation defined..
```suggestion
	// Every AdminNetworkPolicy should have a unique priority value; if two (or more)
	// policies have the same priority and both have rules that match the same connection at the same index (rules have precedence based on ordering), then there is no
	// way for the user to reliably determine which of the policies will actually be
	// applied to the connection. If both policies match different non-overlapping connections
	// then both policies are expected to work correctly
```
something like that? I wanted to make sure we highlight the fact that both ANPs are expected to be created and applied correctly at least for the non overlapping case.. don't want to overengineer this but if I have ANP1 at prio3 whose rule0 matches connectionA and ANP2 at prio3 whose rule1 matches connectionA for sure its guaranteed ANP1 wins, so I guess its a combo of prio+precedence when we speak about two ANPs?

astoycos: > if I have ANP1 at prio3 whose rule0 matches connectionA and ANP2 at prio3 whose rule1 matches connectionA for sure its guaranteed ANP1 wins

Wait Is it? I don't think we can count on ANP _internal_ rule ordering when there's conflicting priorities, i.e ordering rule0 vs rule1 across two different ANPs doesn't make much sense to me 🤔 . In this ^^ case based on what Dan has written I would think the rules do overlap (regardless of rule index) and expect that only one of the ANPs takes effect (which one is up to the implementation). 

But I'm happy to be told otherwise :)

astoycos:
Only thing I might do is consider adding the https://github.com/kubernetes-sigs/network-policy-api/issues/216 like to the comment? I know it's not great to embed issues there but if folks are confused, I think the issue conversation could help clear things up.

nathanjsweet:
LGTM

apis/v1alpha1/adminnetworkpolicy_types.go
npinaeva: I don't think we have ever said how internal precedence works across multiple policies https://github.com/kubernetes-sigs/network-policy-api/blob/main/apis/v1alpha1/adminnetworkpolicy_types.go#L80-L82, so I think we should also add a note there for policies with the same priority, wdyt?

danwinship: > if I have ANP1 at prio3 whose rule0 matches connectionA and ANP2 at prio3 whose rule1 matches connectionA for sure its guaranteed ANP1 wins

No, that's an _allowed_ implementation, but it's not guaranteed.

eg, assuming "ANP1" and "ANP2" are the actual object names, then in the calico-esque "the ANPs get sorted alphabetically" interpretation, then rule1 in ANP1 has precedence over rule0 in ANP2.

And in the "denies beat allows" example, it depends on the actual rules: if ANP1 rule0 is Allow and ANP2 rule1 is Deny, then the result is Deny.

danwinship: how about: "if two (or more) policies with the same priority could both match a connection, then the implementation can apply any of the matching policies to the connection, and there is no way for the user to reliably determine which one it will choose" ?

tssurya: > > if I have ANP1 at prio3 whose rule0 matches connectionA and ANP2 at prio3 whose rule1 matches connectionA for sure its guaranteed ANP1 wins
> 
> No, that's an _allowed_ implementation, but it's not guaranteed.
> 
> eg, assuming "ANP1" and "ANP2" are the actual object names, then in the calico-esque "the ANPs get sorted alphabetically" interpretation, then rule1 in ANP1 has precedence over rule0 in ANP2.
> 
> And in the "denies beat allows" example, it depends on the actual rules: if ANP1 rule0 is Allow and ANP2 rule1 is Deny, then the result is Deny.

ah I see your point; withdrawing that statement then.

PR #233, opened 2024-06-04
Add DomainName field to AdminNetworkPolicyEgressPeer

rahulkjoshi:
https://github.com/kubernetes-sigs/network-policy-api/issues/133


apis/v1alpha1/shared_types.go
```diff
@@ -196,6 +196,21 @@ type AdminNetworkPolicyEgressPeer struct {
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=25
 	Networks []CIDR `json:"networks,omitempty"`
+
```
rahulkjoshi, line 199: @astoycos @npinaeva @tssurya 

Our NPEP says we only support ANP, not BANP. Does that mean we should create two different EgressPeer types for ANP and BANP? 😢

apis/v1alpha1/shared_types.go
```diff
@@ -204,3 +219,25 @@ type AdminNetworkPolicyEgressPeer struct {
...
+//      "my-kubernetes.io", or "wikipedia.org".
+//    - `blog.kubernetes.io` matches only "blog.kubernetes.io".
+//      It does not match "www.kubernetes.io" or "kubernetes.io".
+//    - `*.kubernetes.io` matches subdomains of kubernetes.io.
+//      "www.kubernetes.io", "blog.kubernetes.io", and
+//      "latest.blog.kubernetes.io" match, however "kubernetes.io", and
+//      "wikipedia.org" do not.
+//
+// +kubebuilder:validation:Pattern=`^(\*\.)?([a-zA-z0-9]([-a-zA-Z0-9_]*[a-zA-Z0-9])?\.)+[a-zA-z0-9]([-a-zA-Z0-9_]*[a-zA-Z0-9])?\.?$`
+type DomainName string
```
astoycos, line 243: nit: Newline

apis/v1alpha1/shared_types.go
```diff
@@ -196,6 +196,21 @@ type AdminNetworkPolicyEgressPeer struct {
...
+    // access, DomainNames ALLOW rules should be used with a lower priority
+    // egress deny -- this allows the admin to maintain an explicit "allowlist"
+    // of reachable domains.
+    // 
+    // Support: Extended
+    //
+    // <network-policy-api:experimental>
+    // +optional
+    // +listType=set
+    // +kubebuilder:validation:MinItems=1
```
astoycos, line 212: Do we need a max items here?

apis/v1alpha1/shared_types.go
astoycos: I think we'll have to unfortunately :/ @tssurya can chime in if there's another way since she implemented it in the first place

I dont think it's too big of a deal though, we would just move the `EgressPeer` typedef out of shared_types.go

astoycos (commented):
Small PR, I like it!! Almost there :)

apis/v1alpha1/shared_types.go
astoycos: @rahulkjoshi You should be good to go ahead and make this change :)

astoycos:
Poke @rahulkjoshi Just need some small updated here 👍

apis/v1alpha1/shared_types.go
```diff
@@ -196,6 +196,21 @@ type AdminNetworkPolicyEgressPeer struct {
...
+    DomainNames []DomainName `json:"domainNames,omitempty"`
```
arkadeepsen, line 213:
```suggestion
	
	// DomainNames provides a way to specify domain names as peers.
	// 
	// DomainNames is only supported for ALLOW rules. In order to control
	// access, DomainNames ALLOW rules should be used with a lower priority
	// egress deny -- this allows the admin to maintain an explicit "allowlist"
	// of reachable domains.
	// 
	// Support: Extended
	//
	// <network-policy-api:experimental>
	// +optional
	// +listType=set
	// +kubebuilder:validation:MinItems=1
	DomainNames []DomainName `json:"domainNames,omitempty"`
```
Small nit to match the spacing used by the existing fields.

rahulkjoshi: done, thanks!

apis/v1alpha1/shared_types.go
rahulkjoshi: Added 25 here for now (just to match networks). Because of the wildcard rules it won't effectively limit IPs or even domains matched, but that's probably fine as a first pass.

apis/v1alpha1/shared_types.go
rahulkjoshi: It's done! PTAL

astoycos (approved):
/lgtm
/approve

I think we need https://github.com/kubernetes-sigs/network-policy-api/pull/238 for CI to pass and this to merge but otherwise thanks for all the efforts @rahulkjoshi!

PR #235, opened 2024-06-18
v0.1.6 api review

astoycos:
What type of PR is this?
/kind api-change

What this PR does / why we need it:
This PR is a diff of /apis from v0.1.6 (main branch) to v0.1.1 (release-v0.1.1 branch).

Note: This PR is purely to facilitate review, it is not intended to merge.

Note2: This does not include FQDN support from from @rahulkjoshi in https://github.com/kubernetes-sigs/network-policy-api/pull/233, which will come in our next 0.2.0 release


apis/v1alpha1/adminnetworkpolicy_types.go
```diff
@@ -47,21 +47,35 @@ type AdminNetworkPolicy struct {
...
-	// higher precedence, and are checked before rules with higher priority values.
+	// Priority is a value from 0 to 1000. Policies with lower priority values have
+	// higher precedence, and are checked before policies with higher priority values.
 	// All AdminNetworkPolicy rules have higher precedence than NetworkPolicy or
 	// BaselineAdminNetworkPolicy rules
-	// The behavior is undefined if two ANP objects have same priority.
+	// Every AdminNetworkPolicy should have a unique priority value; if two (or more)
+	// policies with the same priority could both match a connection, then the
+	// implementation can apply any of the matching policies to the connection, and
+	// there is no way for the user to reliably determine which one it will choose.
```
aojea, line 66: There are some implementations that are able to do this deterministically, I did it sorting by name here https://github.com/kubernetes-sigs/kube-network-policies/blob/db089bc6ed6c17a654ef0ea3f3a93d9122041c4e/pkg/networkpolicy/networkpolicyapi.go#L189-L196

I think we can guide users, and say that from the API point of view is difficult to define the exact behavior as it may depend on implementation details, but they should refer to the project implementing this API to understand what is the expected behavior

apis/v1alpha1/adminnetworkpolicy_types.go
```diff
@@ -130,11 +162,16 @@ type AdminNetworkPolicyIngressRule struct {
 // AdminNetworkPolicyEgressRule describes an action to take on a particular
 // set of traffic originating from pods selected by a AdminNetworkPolicy's
 // Subject field.
+// <network-policy-api:experimental:validation>
+// +kubebuilder:validation:XValidation:rule="!(self.to.exists(peer, has(peer.networks) || has(peer.nodes)) && has(self.ports) && self.ports.exists(port, has(port.namedPort)))",message="networks/nodes peer cannot be set with namedPorts since there are no namedPorts for networks/nodes"
```
aojea, line 166: @thockin I think that you mention this to me at one point, about having a way to unit test or validate these expressions

apis/v1alpha1/baselineadminnetworkpolicy_types.go
```diff
@@ -129,17 +161,24 @@ type BaselineAdminNetworkPolicyEgressRule struct {
...
+	// If any AdminNetworkPolicyEgressPeer matches the destination of outgoing
 	// traffic then the specified action is applied.
 	// This field must be defined and contain at least one item.
 	// +kubebuilder:validation:MinItems=1
-	To []AdminNetworkPolicyPeer `json:"to"`
+	// +kubebuilder:validation:MaxItems=100
+	//
+	// Support: Core
+	//
+	To []AdminNetworkPolicyEgressPeer `json:"to"`
```
aojea, line 178: why now the differentiation between Ingress and Egress?

apis/v1alpha1/shared_types.go
```diff
@@ -80,77 +96,111 @@ type Port struct {
...
+	//
 	// +optional
-	NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty"`
-
-	// SameLabels is used to select a set of Namespaces that share the same values
-	// for a set of labels.
-	// To be selected a Namespace must have all of the labels defined in SameLabels,
-	// AND they must all have the same value as the subject of this policy.
-	// If Samelabels is Empty then nothing is selected.
+	Namespaces *metav1.LabelSelector `json:"namespaces,omitempty"`
```
aojea, line 160: much simpler

apis/v1alpha1/baselineadminnetworkpolicy_types.go
rahulkjoshi: The network selector was added to egress (egress controls by CIDR block). We didn't include Ingress yet due to the difficulties reliably determining the source IP of ingress traffic.

https://network-policy-api.sigs.k8s.io/npeps/npep-126-egress-traffic-control/#non-goals

apis/v1alpha1/adminnetworkpolicy_types.go
danwinship: This was hashed out in #216 and #229.

Individual implementations may do something deterministic, but they are not _required_ to. The point of the explanation here is that if you want an "expected behavior", then you have to avoid having two ANPs with the same priority that both match the same packet. (But also, if you _do_ have two ANPs with the same priority that both match the same packet, the "undefined-ness" is much more restricted than it was before; the implementation _must_ obey one of the matching policies. It can't just "always deny", or ignore both policies, etc.)

aojea: I was trying to point out that we can guide the users to ask their implementations about the behaviors, the way is worded now seems that "is impossible to know", when it really means "it is impossible to define a general behavior, but implementations may have a deterministic one"

danwinship: What it really means is "don't do that".

The only scenario that is not well-defined is when you have two ANPs at the same priority that attempt to take different actions on the same packet. No matter what the implementation does, it will fail to obey the administrator's intent, because the administrator has requested that it do two contradictory things. Don't do that. Don't ask which policy it's actually going to use in that case, just fix your policies.

aojea: as a developer is clear, but as a user you can do it if your implementation is deterministic

tssurya:
@astoycos should I help with this? I know last time we spoke we wanted to setup a meeting with going through the PR and hashing out API changes?

astoycos:
For some reason this won't let me re-open now 🤔

astoycos:
Opened https://github.com/kubernetes-sigs/network-policy-api/pull/260 to replace this 🤔 Unless someone can figure out how to re-open we can continue the discussion there

PR #241, opened 2024-07-19
[Policy Assistant] display enhancements

huntergregory:
A few enhancements:
1. display (b)anp rule names in the table and "Flow"
2. make connectivity table easier to read for loopback traffic
3. stop adding an extra row in the explain table when there are no egress policy rules

New flow example:
```
[NPv1] Allow (rule-name)
```

Loopback traffic example (adding "#")
```
simulated connectivity:
+--------+--------+--------+
| TCP/80 | DEMO/A | DEMO/B |
| TCP/81 |        |        |
+--------+--------+--------+
| demo/a | # X    | . X    |
+--------+--------+--------+
| demo/b | . X    | # X    |
+--------+--------+--------+
```


mattfenwick:
I like this -- making those undefined cases explicit!

/lgtm

Issue #247, opened 2024-09-12
[BUG]  AdminNetworkPolicyIngressRule.Ports has two nil values?


fasaxc:
While reviewing our WIP implementation of ANP, I noticed that the Ports field is defined as 
```
	// Ports allows for matching traffic based on port and protocols.
	// This field is a list of ports which should be matched on
	// the pods selected for this policy i.e the subject of the policy.
	// So it matches on the destination port for the ingress traffic.
	// If Ports is not set then the rule does not filter traffic via port.
	//
	// Support: Core
	//
	// +optional
	// +kubebuilder:validation:MaxItems=100
	Ports *[]AdminNetworkPolicyPort `json:"ports,omitempty"`
```
which means that it has two/three "empty" values: 

* `(*[]AdminNetworkPolicyPort)(nil)`
* `&([]AdminNetworkPolicyPort(nil))`
* `&([]AdminNetworkPolicyPort{})`

It's not clear what they're each supposed to mean, "If Ports is not set then the rule does not filter traffic via port".  Clearly a nil pointer is "not set" but what about a pointer to an empty/nil slice?  Would those mean "match no ports" (which in turn is a bit ambiguous since not all protocols have ports)?

I suggest dropping the extra pointer indirection and interpreting any zero-length slice (nil or otherwise) as "do not filter on ports".  If we can't do that, we should at least document the meanings (where I think we should probably make them all mean "do not filter on ports").

tssurya:
cc @astoycos

danwinship:
For future reference, the rationale for the existing semantics is:

- `nil`/not-present _must_ mean "this has no effect on matching". Eg, if we add `Protocols` in the future, then the behavior when you don't specify `Protocols` must be the same as what the behavior was in the version of ANP where that field didn't even exist. eg, "don't restrict based on ICMP". So by extension, not specifying `Ports` should mean "don't restrict based on port".
[...]

PR #250, opened 2024-10-09
Disallow empty port list for ANP peers.

npinaeva:
Based on discussion https://github.com/kubernetes-sigs/network-policy-api/issues/247
Zero-length port list is different from the nil port list, and it means do not match any port, which makes that peer not matching any traffic. 
This is confusing and is more of an accident than an intention.

### DIFF
--- a/apis/v1alpha1/adminnetworkpolicy_types.go
+++ b/apis/v1alpha1/adminnetworkpolicy_types.go
@@ -155,6 +155,7 @@ type AdminNetworkPolicyIngressRule struct {
 	// Support: Core
 	//
 	// +optional
+	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=100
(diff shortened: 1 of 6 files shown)


aojea:
@danwinship don't you bump the versions for these changes? v1alpha2?

danwinship:
/approve

> @danwinship don't you bump the versions for these changes? v1alpha2?

pfft. YOLO!

Yeah, it's technically an API break, but it's alpha. I guess people who are shipping this alpha API to customers (*meaningful glance*) will have to decide whether or not to ship the break.

(leaving un-lgtm'ed to give others a chance to comment)

npinaeva:
had to rebase to fix CI

aojea:
I see, ok, let me know when you have planned a release to update the network policies,

aojea:
we may want to do a release before kubecon, and then freeze there, otherwise we loose credibility too

aojea:
we can break between alpha minors

PR #258, opened 2024-10-22
OWNERS: Add aojea as an approver

npinaeva:
cc @aojea @danwinship

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,6 +1,7 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 approvers:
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -4,3 +4,4 @@ approvers:
   - astoycos
   - Dyanngg
   - danwinship
+  - aojea
```
danwinship, line 7: can you fix this to be alphabetical?

npinaeva: never commit late in the evening :)

Issue #255, opened 2024-10-21
[Policy Assistant] Cut the first release


huntergregory:
_EDIT: we are going to use `policy-assistant` instead of `pola` per #262_ 

Perhaps we can cut a release just for policy assistant, called v0.0.1-pola? Looks like there are no overlapping or bad connotations for this POLA abbreviation from my Google search 😅 

### Requirements

- CI/CD:
  - #170 
  - #254
  - #246
  - cut release
- features:
  - #245
  - #220

### Nice to have

- ~~#212~~:
  - added warning in #261
- ~~#151~~
  - removed unimplemented CLI args per #245

huntergregory:
Discussed cutting a `v0.0.1-pola` release for the `pola` CLI in last week's meeting. We will target a release date of November 5th

huntergregory:
Discussed tying pola version to the API version used. Discussed how the first release might not follow that format though.

Two ideas for first release:
- idea 1: `v0.0.1-pola`
- idea 2: `v0.1.1-pola-20241105`
  - `<api-version-used>-pola-<date>`

npinaeva:
One question I have is do we plan to support multiple API versions with the same binary? It sounds like it could be useful to not reinstall the binary if you have 2 clusters with different ANP versions. In that case the tool could have its own versions, just adding support for newer API versions as it progresses?

huntergregory:
> One question I have is do we plan to support multiple API versions with the same binary? It sounds like it could be useful to not reinstall the binary if you have 2 clusters with different ANP versions. In that case the tool could have its own versions, just adding support for newer API versions as it progresses?

After revisiting this question, I'm not sure we can support multiple API versions in the same CLI: https://stackoverflow.com/questions/63254746/how-to-import-two-versions-of-the-same-go-module

huntergregory:
release complete: https://github.com/kubernetes-sigs/network-policy-api/releases/tag/v0.0.1-policy-assistant

PR #264, opened 2024-11-05
[Policy Assistant] Add support for dynamic workload traffic via POLA cli

gabrielggg:
Hey @huntergregory , as discussed, here is the pr for allowing dynamic traffic via cli. This is to finish https://github.com/kubernetes-sigs/network-policy-api/issues/220 . And to allow https://github.com/kubernetes-sigs/network-policy-api/issues/255 .

To test you cand do something like this:

image

Please check it out.


cmd/policy-assistant/pkg/cli/analyze.go
```diff
@@ -296,7 +308,10 @@ func shouldIncludeANPandBANP(client *kubernetes.Clientset) (bool, bool) {
 	return includeANP, includeBANP
 }
 
-func VerdictWalkthrough(policies *matcher.Policy) {
+func VerdictWalkthrough(policies *matcher.Policy, sourceWorkloadTraffic string, destinationWorkloadTraffic string, port int, protocol string) {
```
huntergregory, line 311: can we print a warning and exit the function for bad input?

I guess that would entail:
1. either traffic peer not found
2. `v1.Protocol(protocol)` not equal to TCP, UDP, SCTP

cmd/policy-assistant/pkg/cli/analyze.go
```diff
@@ -305,31 +320,52 @@ func VerdictWalkthrough(policies *matcher.Policy) {
...
-			PodLabels:       map[string]string{"pod": "b"},
-			NamespaceLabels: map[string]string{"kubernetes.io/metadata.name": "demo"},
-			Namespace:       "demo",
+			PodLabels:       destinationWorkloadInfo.Internal.PodLabels,
+			NamespaceLabels: destinationWorkloadInfo.Internal.NamespaceLabels,
+			Namespace:       destinationWorkloadInfo.Internal.Namespace,
+			Workload:        destinationWorkloadInfo.Internal.Workload,
 		},
-		IP: "10.0.0.5",
+		IP: destinationWorkloadInfo.Internal.Pods[0].IP,
```
huntergregory, line 359: Could we do nil checks/empty list check to avoid chance of panic?

huntergregory (changes requested):
Thank you for the effort last minute @gabrielggg !

This looks great. If you could address the minor suggestions today that'd be amazing. Could you test each of the 5 workload types and when `--port` and `--protocol` are unspecified vs. specified?

Would be awesome if we could also support specifying a _traffic.json_ with Pod labels etc. in case the user doesn't have a cluster and can't specify workload name. It also could be nice to support specifying workload in the _traffic.json_, but that is lower priority if we run out of time.

Let me know your thoughts! Thanks!

gabrielggg:
hey @huntergregory, thanks for the review , i just committed some code to at least fix the conversations you opened, the next commit will include support for the traffic.json file, but i think that will be ready for the night or maybe tomorrow.

gabrielggg:
Hey @huntergregory , it's working now, and supporting cli arguments and also traffic.json file including native workloads on the file, please check it out.
image

cmd/policy-assistant/pkg/cli/analyze.go
```diff
@@ -296,59 +308,291 @@ func shouldIncludeANPandBANP(client *kubernetes.Clientset) (bool, bool) {
...
+
+		if protocol != "TCP" && protocol != "UDP" && protocol != "SCTP" {
+			logrus.Fatalf("Bad Protocol Value: protocols supported are TCP, UDP and SCTP")
+		}
+
+		sourceWorkloadInfo = matcher.WorkloadStringToTrafficPeer(sourceWorkloadTraffic)
+		destinationWorkloadInfo = matcher.WorkloadStringToTrafficPeer(destinationWorkloadTraffic)
+
+		if sourceWorkloadInfo.Internal.Pods == nil || destinationWorkloadInfo.Internal.Pods == nil {
+			return
```
huntergregory, line 558: could we `logrus.Warn()` here that we couldn't find source workload or couldn't find dst workload?

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -59,26 +59,95 @@ func labelsToString(labels map[string]string) string {
...
+	var dst string
 	if t == nil || t.Source == nil || t.Destination == nil {
 		return "<undefined>"
 	}
 
-	src := t.Source.Internal.Workload
-	if src == "" {
-		if t.Source.Internal == nil {
-			return "<undefined>"
+	if t.Source.Internal == nil && t.Destination.Internal == nil {
```
huntergregory, line 68: src string should be independent of dst string right? Could we refactor this to be something like this?
```go
var src string
if t.Source != nil {
    src := t.Source.Internal.PrettyString()
}
var dst string
if t.Destination != nil {
    dst := t.Destination.Internal.PrettyString()
}
return fmt.Sprintf("%s -> %s:%d (%s)", src, dst, t.ResolvedPort, t.Protocol)
 ```

File: cmd/policy-assistant/pkg/cli/analyze.go [hunk 5]
```diff
@@ -296,59 +308,291 @@ func shouldIncludeANPandBANP(client *kubernetes.Clientset) (bool, bool) {
...
+
+	if trafficPath != "" && (sourceWorkloadTraffic != "" || destinationWorkloadTraffic != "" || port != 0 || protocol != "") {
+		logrus.Fatalf("%+v", errors.Errorf("If using traffic path, you can't input traffic via CLI and viceversa"))
+	}
+
+	if trafficPath != "" {
+		allTraffics, err := json.ParseFile[[]*matcher.Traffic](trafficPath)
+		utils.DoOrDie(err)
+		for _, traffic := range *allTraffics {
+			if traffic.Source.Internal == nil && traffic.Destination.Internal == nil {
```
huntergregory, line 324: podA should be independent of podB right? Could we refactor to something like this? Also adding a couple suggestions for input checks in below code
```go
for _, traffic := range *allTraffics {
    if traffic.Source == nil || traffic.Destination == nil {
        logrus.Warn(...)
        continue
    }

    // also warn/continue if resolvedPort and protocol are invalid
    
    podA := populateTrafficPeer(traffic.Source.Internal)
    podB := populateTrafficPeer(traffic.Destination.Internal)
    allTrafficTmp := matcher&.Traffic{
	    Source:       podA,
	    Destination:  podB,
	    ResolvedPort: traffic.ResolvedPort,
	    Protocol:     traffic.Protocol,
    }
    allTraffic = append(allTraffic, allTrafficTmp)
}
```

huntergregory (changes requested):
Thanks @gabrielggg that result looks amazing! I appreciate you going above and beyond supporting the whole traffic.json in this PR. I had minor suggestions for validating input. I also feel like having all these if-else statements could be bug prone 😕. Luckily, it looks like all the branches have shared logic that can be refactored into a couple functions? I added comments with outlines for much smaller code.

I still would love to release this change for the upcoming KubeCon, but I feel like I unfortunately can't do a just review of the code if there are all the if-else branches

gabrielggg:
hey @huntergregory , thanks for the advice, i finished with the refactoring! Please check it out!

cmd/policy-assistant/pkg/cli/analyze.go
gabrielggg: This is allready handled on the translate func, if the workload doesnt exist or if it exists but is scaled down to 0 i will display this : 
image

cmd/policy-assistant/pkg/cli/analyze.go
```diff
@@ -296,59 +308,114 @@ func shouldIncludeANPandBANP(client *kubernetes.Clientset) (bool, bool) {
...
 }
 
-func VerdictWalkthrough(policies *matcher.Policy) {
+func VerdictWalkthrough(policies *matcher.Policy, sourceWorkloadTraffic string, destinationWorkloadTraffic string, port int, protocol string, trafficPath string) {
+	var sourceWorkloadInfo matcher.TrafficPeer
+	var destinationWorkloadInfo matcher.TrafficPeer
+	var allTraffic []*matcher.Traffic
+
+	if trafficPath != "" && (sourceWorkloadTraffic != "" || destinationWorkloadTraffic != "" || port != 0 || protocol != "") {
+		logrus.Fatalf("%+v", errors.Errorf("If using traffic path, you can't input traffic via CLI and viceversa"))
```
huntergregory, line 317: Thinking about how a user will know what parameters to use for this mode. Think you solved the first problem of specifying both traffic path and other args.

Could we also add a check up here that if traffic path is not set, then all the other args must be set.

I would also word this existing check and the new check to explicitly write out the arg names because there's currently no place where the user would learn which args need to be set for "walkthrough" mode. Something like: "for this mode, must either set --traffic-path or set all of --source-workload --dst-workload..."

cmd/policy-assistant/pkg/cli/analyze.go
```diff
@@ -296,59 +308,114 @@ func shouldIncludeANPandBANP(client *kubernetes.Clientset) (bool, bool) {
...
+
+			// Determine source and destination peer information
+			sourceInternal := traffic.Source.Internal
+			destinationInternal := traffic.Destination.Internal
+
+			podA = matcher.CreateTrafficPeer(traffic.Source.IP, nil)
+			podB = matcher.CreateTrafficPeer(traffic.Destination.IP, nil)
+
+			// Resolve internal peer info based on workloads
+			//sourcePeer := GetInternalPeerInfo(sourceInternal.Workload)
```
huntergregory, line 334: will we need this? can we delete any new unused code?

cmd/policy-assistant/pkg/cli/analyze.go
```diff
@@ -296,59 +308,114 @@ func shouldIncludeANPandBANP(client *kubernetes.Clientset) (bool, bool) {
...
+					PodLabels:       destinationInternal.PodLabels,
+					NamespaceLabels: destinationInternal.NamespaceLabels,
+					Namespace:       destinationInternal.Namespace,
+					Workload:        destinationInternal.Workload,
+				})
+			}
+
+			// Special case handling for workload-specific traffic (internal vs. external)
+			if sourceInternal != nil && destinationInternal != nil {
+				if sourceInternal.Workload != "" && destinationInternal.Workload != "" {
```
huntergregory, line 358: nit: src and dst can be independent again. could move this check into
```go
if sourceInternal != nil {
    if sourceInternal.Workload != "" {
        ....
    } else {
        ....
    }
}
```

cmd/policy-assistant/pkg/cli/analyze.go
```diff
@@ -296,59 +308,114 @@ func shouldIncludeANPandBANP(client *kubernetes.Clientset) (bool, bool) {
...
+	} else {
+
+		if protocol != "TCP" && protocol != "UDP" && protocol != "SCTP" {
+			logrus.Fatalf("Bad Protocol Value: protocols supported are TCP, UDP and SCTP")
+		}
+
+		sourceWorkloadInfo = matcher.WorkloadStringToTrafficPeer(sourceWorkloadTraffic)
+		destinationWorkloadInfo = matcher.WorkloadStringToTrafficPeer(destinationWorkloadTraffic)
+
+		if sourceWorkloadInfo.Internal.Pods == nil || destinationWorkloadInfo.Internal.Pods == nil {
```
huntergregory, line 380: if translate fails, will `Internal` be nil? Just wondering in case of nil pointer panic

cmd/policy-assistant/pkg/matcher/traffic.go
```diff
@@ -58,29 +58,32 @@ func labelsToString(labels map[string]string) string {
 	return strings.Join(slice.Map(format, slice.Sort(maps.Keys(labels))), "\n")
 }
 
-func (t *Traffic) PrettyString() string {
-	if t == nil || t.Source == nil || t.Destination == nil {
-		return "<undefined>"
+// Helper function to generate the string for source or destination
+func (t *Traffic) formatPeer(peer *TrafficPeer) string {
```
huntergregory, line 62: nit:
```suggestion
func formatPeer(peer *TrafficPeer) string {
```

huntergregory (changes requested):
Thanks @gabrielggg, on the home stretch. I'd like to make one more change for the user experience. Good to merge after! We are planning to cut the release last minute on Monday after 12:00 PST if you could possibly update before then (the conference starts Tuesday). Would that work?

cmd/policy-assistant/pkg/cli/analyze.go
huntergregory: `len(pods) > 0` would be safer in case the code was ever modified to return an empty list for Pods

cmd/policy-assistant/pkg/matcher/traffic.go
gabrielggg: Can't do this: 
image

cmd/policy-assistant/pkg/cli/analyze.go
gabrielggg: Internal won´t ne nill: 
image

example scenario:
image

gabrielggg:
Hey @huntergregory , thanks again for the review, i tackled all your suggestions and also squashed the commits. Please check it out!

huntergregory:
Looks good at a glance. Adding approval to hopefully kick off the GitHub action to make sure it passes

/approve

Will add a LGTM hopefully tomorrow morning

huntergregory:
Trying the other way around. Lgtm first to kickoff GH action

/lgtm

cmd/policy-assistant/pkg/cli/analyze.go
```diff
@@ -296,59 +308,113 @@ func shouldIncludeANPandBANP(client *kubernetes.Clientset) (bool, bool) {
...
+					NamespaceLabels: destinationInternal.NamespaceLabels,
+					Namespace:       destinationInternal.Namespace,
+					Workload:        destinationInternal.Workload,
+				})
+			}
+
+			// Special case handling for workload-specific traffic (internal vs. external)
+			if sourceInternal != nil {
+				if sourceInternal.Workload != "" {
+					podA = matcher.GetInternalPeerInfo(sourceInternal.Workload)
```
huntergregory, line 357: small bug: if workload not found, `continue`

cmd/policy-assistant/pkg/cli/analyze.go
```diff
@@ -96,6 +104,10 @@ func SetupAnalyzeCommand() *cobra.Command {
 	command.Flags().StringVar(&args.TrafficPath, "traffic-path", "", "path to json traffic file, containing of a list of traffic objects")
 	command.Flags().StringVar(&args.ProbePath, "probe-path", "", "path to json model file for synthetic probe")
 	command.Flags().DurationVar(&args.Timeout, "kube-client-timeout", DefaultTimeout, "kube client timeout")
+	command.Flags().StringVar(&args.SourceWorkloadTraffic, "source-workload-traffic", "", "Source workload traffic in this form namespace/workloadType/workloadName")
```
huntergregory, line 107: I would recommend renaming these two lines to `--source-workload` and `--destination-workload` for less typing?

cmd/policy-assistant/pkg/matcher/traffic.go
huntergregory: I think you could remove the `t.` but this is just a style nit and not a blocker

huntergregory (commented):
Thanks @gabrielggg. Ready to approve but would prefer if we could fix the minor bug, and I'd recommend renaming the args, but neither is a blocker

cmd/policy-assistant/pkg/cli/analyze.go
huntergregory: or even `--src-workload`, `dst-workload`?

cmd/policy-assistant/pkg/cli/analyze.go
gabrielggg: If i put continue there it breaks the loop and doesn't append the traffic, to the array, i tested it and it filtered all traffic where there wasnt a workload. On the other hand, i modified the translate func to return more info of the workload not founded.

cmd/policy-assistant/pkg/cli/analyze.go
gabrielggg: good idea!

huntergregory (approved):
Thanks @gabrielggg ! This is a great feature

/lgtm
/approve

PR #273, opened 2024-11-15
test: add retry for anp/banp connection test

hongzhen-ma:
add retry for anp/banp connection test since the policy need some time to take effect but the tests run check just once, which will lead to unnecessary failure

part of #108

### DIFF
--- a/conformance/utils/kubernetes/helper.go
+++ b/conformance/utils/kubernetes/helper.go
@@ -74,14 +74,35 @@ func PokeServer(t *testing.T, client k8sclient.Interface, kubeConfig *rest.Confi
 		protocolArg,
 		ipPortArg), " ")
 
-	stdout, stderr, err := RunCommandFromPod(client, kubeConfig, clientNamespace, clientPod, connectCommand)
-	// TODO(tssurya): See if we need to add a wait&retry mechanism to test connectivity
-	// See https://github.com/kubernetes-sigs/network-policy-api/issues/108 for details.
-	if err != nil && stderr == "" {
-		// If err != nil and stderr == "", then it means this probe failed because of the command instead of connectivity.
-		t.Logf("FAILED to execute command %s on pod %s/%s: %v", connectCommand, clientNamespace, clientPod, err.Error())
+	var stdout, stderr string
+	var err error
+	// reuse the RequestTimeout parameter as context timeout
+	waitErr := wait.PollUntilContextTimeout(context.Background(), timeout, 3*timeout, true, func(ctx context.Context) (bool, error) {
+		stdout, stderr, err = RunCommandFromPod(client, kubeConfig, clientNamespace, clientPod, connectCommand)
+		// See https://github.com/kubernetes-sigs/network-policy-api/issues/108 for details.
+		if err != nil && stderr == "" {
+			// If err != nil and stderr == "", then it means this probe failed because of the command instead of connectivity.
+			t.Logf("FAILED to execute command %s on pod %s/%s: %v", connectCommand, clientNamespace, clientPod, err.Error())
+			return false, err
+		}
+
+		if shouldConnect && len(stderr) == 0 {
+			return true, nil
+		} else if !shouldConnect && strings.Contains(stderr, "TIMEOUT") {
+			return true, nil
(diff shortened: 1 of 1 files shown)


aojea:
is this needed? I didn't see the test flake in the CI and adding retries for network policies means the pods will not be protected during that time, that is something bad for the users

hongzhen-ma:
> is this needed? I didn't see the test flake in the CI and adding retries for network policies means the pods will not be protected during that time, that is something bad for the users

is this needed?
--- We did encounter this problem in e2e test for Kube-OVN, so we want to add a retry. After add retry, the result succeed in our e2e test.
Of course, we will also look at the problems in our implementation and try to solve it.

Another point is that retry is about connection results after an anp or a banp applied, and does not modify policy rules during
the retry check process.

hongzhen-ma:
> > Of course, we will also look at the problems in our implementation and try to solve it.
> 
> Modifying tests that may hide problems lowers the quality bar ... the tests are done to assert on behaviors not to just pass,

Thanks for the reply.
We will troubleshoot implementation issues firstly.

fasaxc:
FWIW, I've seen some flakes on Calico too.  Haven't analysed them in detail yet but there are classes of updates that can race with connectivity checks.  Calico's model is as follows:

* All connectivity is blocked by default.
* Calico learns about the pod, programs the current known policy for that pod atomically.
* When policy or selector matches change, Calico updates the dataplane incrementally.  We make updates in a way that avoids flapping to open.

Some examples of races with that model:

* If you have an allow rule and switch it to deny, an immediate connectivity check might still see allow but after a few ms it should go to deny and stay that way.
* Similarly, if you have an allow rule and you add a higher precedence deny.
* If you have an "deny to selector" rule and you add a new pod that matches the selector, it takes some short time to add the pod to the deny list in the dataplane (which is why deny to selector rules are sub-optimal!).  After adding a pod, you'll need to retry the check to avoid races.

I expect that `kube-network-policy` has the lowest latency in these areas because it only needs to get the update to user-space, not program the kernel too.

aojea:
We discussed this in slack, and just to clarify, and apoligize if this came wrong, I'm not advocating for zero latency, but just before bumping timeouts to have a better understanding on the problem, it most probably bias, as it happens a lot to me that the immediate solution for a flake is adding a sleep or a retry, and a considerable number of times there is a bug that we ignore.

In in k/k there is 10 s IIUIC https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/netpol/kubemanager.go#L328-L350

I think ANP has 3 seconds for the tests https://github.com/kubernetes-sigs/network-policy-api/blob/b1d9608f0c8b9b799d3af0053374317992bf0579/conformance/utils/config/timeout.go#L50-L58

I'm +1 on adapting the tests to allow 10s for the progamming latency as it is the same we have in k/k, but I really think we should not go beyond that , sounds fair?

/reopen

hongzhen-ma:
> In in k/k there is 10 s IIUIC https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/netpol/kubemanager.go#L328-L350

Thanks for the information.
I will first look at the implementation of netpol test case, and then come back to see how to modify this test

fasaxc:
@aojea yep, 100% agree.  Calico's FV tests also use [10s timeout](https://github.com/projectcalico/calico/blob/master/felix/fv/connectivity/conncheck.go#L297) for connectivity/policy convergence.

I think the best strategy I've found for these kinds of tests is to measure convergence time (with a a very long abort timeout) and then assert that the time taken was less than a "reasonable" time.  

Then you get one of a three results instead of just pass/fail:

* Pass: converged and in the expected time
* Fail: converged but it took 30s instead of 10s
* Fail: did not converge, even after `<long timeout>`.

Being able to distinguish "converged but took too longer than we want" from "did not converge after an unreasonably long time" makes the debugging a lot easier.

aojea:
> Then you get one of a three results instead of just pass/fail:
> 
> * Pass: converged and in the expected time
> * Fail: converged but it took 30s instead of 10s
> * Fail: did not converge, even after `<long timeout>`.

I love that

Issue #230, opened 2024-05-21
[ENHANCEMENT] add dry-run mode for ANP/BANP


npinaeva:
**Is your enhancement request related to a problem? Please describe.**
Users may want to ensure no unexpected connections will be denied by a new (B)ANP. 
A dry-run mode should not drop any connections, and provide a way to get feedback from the networking plugin (e.g. logging) to see which connections will be dropped once this (B)ANP is enforced.

**Describe the solution you'd like**
A potential solution if to have a dry-run flag, that would turn `Deny` action into `Allow`+log (log meaning any type of feedback from the networking plugin; turn into means internally, not on a CR level). As `Deny` action would immediately drop the traffic, `Allow` immediately allows it without any further policies evaluation, so the result should be accurate.

**Describe alternatives you've considered**
- Use Policy Assistant https://github.com/kubernetes-sigs/network-policy-api/issues/221, may have some limitations. Needs additional discussion.
- Leave it to be implementation-specific. More details on why it may be difficult to implement need to be provided.

**Additional context**
First discussed on May 21, 2024 in the SIG meeting.

huntergregory:
> Users may want to ensure no unexpected connections will be denied by a new (B)ANP.

This is definitely an important scenario for user tooling. In March at KubeCon ([recording](https://youtu.be/riSv0g-TNtI?si=6gxDON4jK2xf4n-e&t=1239), [slides](https://static.sched.com/hosted_files/kccnceu2024/bc/KubeConEU24_NetworkPolicy_API.pptx.pdf)), we proposed that upstream tooling would solve this scenario, and there was good excitement and feedback which was the catalyst for putting #221 (same scenario) on the roadmap for Policy Assistant.

I think we should solve this scenario in upstream tooling (Policy Assistant) as opposed to a dry-run mode downstream. A few reasons:
- ***Our upstream tooling handles this and more. It is already a “dry-run” policy engine.***
- Benefits of upstream vs. downstream tooling:
    - Developing new features will be quicker/easier upstream.
    - One set of changes will benefit all users regardless of implementation.
    - Less work for implementations.
- ***I think upstream should provide a consistent story around user tooling. If upstream suggests using an upstream tool for one scenario, and a downstream tool/configuration for another scenario, there is a disjointed experience for users, and mixed messaging.***

There were two concerns I recall from yesterday’s discussion around Policy Assistant’s solution to this scenario:
- How could Policy Assistant know about connections to determine denied connections?
[...]

PR #276, opened 2024-11-19
NPEP-230: Add dry-run mode.

npinaeva:
Tracking issue https://github.com/kubernetes-sigs/network-policy-api/issues/230

### DIFF
--- a/npeps/npep-230-dry-run.md
+++ b/npeps/npep-230-dry-run.md
@@ -1,0 +1,75 @@
+# NPEP-95: NPEP template
+
+* Issue: [#230](https://github.com/kubernetes-sigs/network-policy-api/issues/230)
+* Status: Provisional
+
+## TLDR
+
+Add dry-run mode for (B)ANP to allow "disabling" policies without deleting them.
+This should allow implementations add extra audit/monitoring/logging capabilities on top of this feature.
+
+## Goals
+
+A (B)ANP dry-run mode should not affect any connections, i.e. it should be treated as if did not exist.
+Networking plugins can interpret the contents of the object to provide feedback (e.g. via logging or observability tools) 
+to see which connections will be dropped/allowed once this (B)ANP is enforced.
+
+## Non-Goals
+
+Define exact logging/observability formats for network plugins.
+
+## Introduction
+
+Users may want to ensure no unexpected connections will be denied by a new (B)ANP.
+A dry-run mode should not affect any connections, and allow the networking plugin to provide feedback 
+(e.g. via logging or observability tools) to see which connections will be dropped/allowed once this (B)ANP is enforced.
+
+## User-Stories/Use-Cases
+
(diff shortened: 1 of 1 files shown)


npeps/npep-230-dry-run.md
```diff
@@ -0,0 +1,62 @@
...
+* Status: Provisional
+
+## TLDR
+
+Add dry-run mode for (B)ANP to allow "disabling" policies without deleting them.
+This should allow implementations add extra audit/monitoring/logging capabilities on top of this feature.
+
+## Goals
+
+A dry-run mode should not affect any connections, and allow the networking plugin to provide feedback
```
bowei, line 13: Let's be more clear on this:

- A NetworkPolicy object in dry-run mode will not enforce any K8s network policy semantics e.g. it should be treated as if it did not exist. Controllers are free to interpret the contents of the object for purposes of observability/debugging however.

npeps/npep-230-dry-run.md
```diff
@@ -0,0 +1,62 @@
+# NPEP-95: NPEP template
+
+* Issue: [#230](https://github.com/kubernetes-sigs/network-policy-api/issues/230)
+* Status: Provisional
+
+## TLDR
+
+Add dry-run mode for (B)ANP to allow "disabling" policies without deleting them.
```
bowei, line 8: Why is this not also a proposal for NetworkPolicy itself?

npeps/npep-230-dry-run.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Alternatives
+
+- Use Policy Assistant https://github.com/kubernetes-sigs/network-policy-api/issues/221, may have some limitations. Needs additional discussion.
+- Leave it to be implementation-specific. This required a copy of ANP CRD that changes behaviour (doesn't really apply) with dry-run flag.
+
+## References
+
+Similar features:
+- calico: https://docs.tigera.io/calico-cloud/network-policy/staged-network-policies
```
fasaxc, line 60: I talked to the folks that designed Calico's staged policy and I asked why we added a second resource type `StagedNetworkPolicy` instead of an "enforced" flag.  The answer was that it allows for staging a _change_ to an existing policy as well as staging a completely new policy.  You can have a staged network policy with name "foo" and order 123 as well as an unstaged one with the same name and order.  Then the staged version should replace the active one when simulating the upcoming policy verdict.

That's not quite the same as creating a second policy at the same order with a different name; it wouldn't be clear if the user is asking us to simulate both policies being present or only the new one.

npinaeva: I think this is a valuable use case, added a section about this to NPEP.
I also think it may be significantly more difficult to implement as it requires some simulation engine (at least for ovn-kubernetes), would be nice to gather more feedback for other implementations.

bowei:
The question here is almost philosophical -- does it make sense to do something small that is universal but limited or go for something more fully-featured but probably will not be able to be implemented everywhere.

Issue #285, opened 2025-02-12
[ENHANCEMENT] Combine ANP/BANP into one CRD


fasaxc:
**Is your enhancement request related to a problem? Please describe.**

This came up in the API review with @thockin.  He suggested we look again at that split and consider how the API might evolve; in particular, what if we remove the singleton aspect from BANP and it ends up the same as ANP (with only tiny gotcha-level differences).

**Describe the solution you'd like**

Represent ANP and BANP with one CRD type and represent the difference using some other primitive. 

My favoured approach would be to add a `tier` field to the combined ANP type that can be set to `<value that makes it an ANP>` `<value that makes it a BANP>`.  This has the advantage that it's extensible to new "tiers" later.

TBD: do we immediately remove the BANP singleton restriction and add priority, or do we defer that (validate against it?)

**Describe alternatives you've considered**

On the call we discussed alternatives:

* Tier field.
* Negative priority for ANP, positive for BANP.

**Additional context**

Calico already has the concept of tiered policy; it is the feature that we use to implement ANP and BANP internally so there's precedent for the name and the feature has been popular with our enterprise users.  Calico allows user defined tiers, which may be something to consider for the future, but I think we can extend in that way later if required.

PR #289, opened 2025-03-03
npep-285: Combine ANP and BANP

npinaeva:
Current version doesn't suggest a specific API, but outlines different options. After this EP gets some feedback, it will be updated with the winning options.
tracking issue https://github.com/kubernetes-sigs/network-policy-api/issues/285

### DIFF
--- a/npeps/npep-285-combine-crds.md
+++ b/npeps/npep-285-combine-crds.md
@@ -1,0 +1,100 @@
+# NPEP-285: NPEP template
+
+* Issue: [#285](https://github.com/kubernetes-sigs/network-policy-api/issues/285)
+* Status: Provisional
+
+## TLDR
+
+Combine ANP and BANP into a single CRD.
+
+## Goals
+
+- Outline existing differences between ANP and BANP.
+- Merge ANP and BANP into a single CRD.
+
+## Non-Goals
+
+(What is explicitly out of scope for this proposal.)
+
+## Introduction
+
+ANP and BANP are 2 layers of policies around NetworkPolicy, evaluated as shown in the diagram below.
+
+![image](../site-src/images/ANP-api-model.png)  
+Currently, ANP and BANP are 2 separate CRDs with the following differences:
+- BANP is a singleton, which is also the reason why it doesn't have `spec.priority` field.
+- ANP supports `Pass` action, while BANP doesn't. `Pass` action could be considered a delegation to the NetworkPolicy, 
+or it could be seen as return from the current policy layer .
+- ANP supports `domainNames` matching for egress rules, while BANP doesn't.
+
+## User-Stories/Use-Cases
+
+Story 1: Multiple BANPs
+
+As a cluster admin, I want to apply different BANPs to different sets namespaces, so that I can have different default trust levels.
+For example, "system" namespaces, like kube-system, could have a less strict BANP than "user" namespaces.
+
+Story 2: Single CRD
+
+As a cluster admin, I prefer to have 1 CRD for both ANP and BANP, so that it is easier to manage, understand, and monitor.
+As an API implementation developer, I prefer to have 1 CRD for both ANP and BANP, so that I can reduce the complexity of the code.
+As an API developer, I prefer to have 1 CRD for both ANP and BANP, so that I can reduce the complexity of the API and maintenance load.
+
+## API
+
+### Combining ANP and BANP
+
+To combine ANP and BANP, we need to have a way to specify policy layer. Currently, we have the following options:
+- Add a new field `spec.tier` to distinguish ANP vs BANP. The values could be `Overrride` and `Baseline`.
+- Change `spec.priority` field to have a pre-defined value for NetworkPolicy, use numbers below for ANP, and numbers above for BANP.
+For example, if we set NetworkPolicy priority to 0, then negative priorities would mean ANP, and positive priorities would mean BANP. 
+
+#### Find a better name for `spec.priority`
+
+While we are here, `priority` field name was brought to our attention multiple times as confusing. It should be mostly related to the fact
+that "higher priority" means "lower number", which is counter-intuitive. Suggested alternatives are "order", "precedence", "sequence".
+
+### Resolving differences between ANP and BANP
+
+1. BANP is a singleton, which is also the reason why it doesn't have `spec.priority` field.  
+Make BANP non-singleton and apply `spec.priority` field to it.
+
+2. ANP supports `Pass` action, while BANP doesn't. `Pass` action could be considered a delegation to the NetworkPolicy,
+  or it could be seen as return from the current policy layer.   
+We have 2 options: allow using `Pass` for BANP, meaning that the rest od BANPs will be skipped or forbid this action for BANP.  
+Unless we find strong reasons to forbid `Pass` for BANP, it makes sense to allow it to make the API more uniform.
+
+3. ANP supports `domainNames` matching for egress rules, while BANP doesn't.  
+The original reasoning from the 
+[FQDN NPEP](https://github.com/kubernetes-sigs/network-policy-api/blob/feae59d056cf87131338a8449d31b85dc1d9790f/npeps/npep-133-fqdn-egress-selector.md?plain=1#L33-L38)
+is
+> Since Kubernetes NetworkPolicy does not have a FQDN selector, adding this
+capability to BaselineAdminNetworkPolicy could result in writing baseline
+rules that can't be replicated by an overriding NetworkPolicy. For example,
+if BANP allows traffic to `example.io`, but the namespace admin installs a
+Kubernetes Network Policy, the namespace admin has no way to replicate the
+`example.io` selector using just Kubernetes Network Policies.
+
+DNS Name selector is only [supported](https://github.com/kubernetes-sigs/network-policy-api/blob/feae59d056cf87131338a8449d31b85dc1d9790f/apis/v1alpha1/adminnetworkpolicy_types.go#L282)
+with the `Allow` rule, which means the desired override by the NetworkPolicy could only be `Deny`. 
+Due to the default-deny nature of NetworkPolicy, all required traffic in a namespace should be explicitly allowed, which means
+the override of the BANP `Allow` DNS Name selector will happen automatically.
+
+## Conformance Details
+
+(This section describes the names to be used for the feature or
+features in conformance tests and profiles.
+
+These should be `CamelCase` names that specify the feature as
+precisely as possible, and are particularly important for
+Extended features, since they may be surfaced to users.)
+
+## Alternatives
+
+The only alternative is to leave ANP and BANP as 2 CRDs, the main advantage on this approach is that implementations that already support
+an alpha version of the API don't need to introduce major changes. (But we have alpha to allow big changes like this, right?)
+
+## References
+
+Was discussed in the [Network Policy API meeting](https://docs.google.com/document/d/1AtWQy2fNa4qXRag9cCp5_HsefD7bxKe3ea2RPn8jnSs) on 11th and 25th February 2025.
+


npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,100 @@
...
+As a cluster admin, I prefer to have 1 CRD for both ANP and BANP, so that it is easier to manage, understand, and monitor.
+As an API implementation developer, I prefer to have 1 CRD for both ANP and BANP, so that I can reduce the complexity of the code.
+As an API developer, I prefer to have 1 CRD for both ANP and BANP, so that I can reduce the complexity of the API and maintenance load.
+
+## API
+
+### Combining ANP and BANP
+
+To combine ANP and BANP, we need to have a way to specify policy layer. Currently, we have the following options:
+- Add a new field `spec.tier` to distinguish ANP vs BANP. The values could be `Overrride` and `Baseline`.
```
bowei, line 48: I think it probably makes more sense to have a separate tier vs try to sort things by priority. The issue is that baseline never applies if the user has a network policy defined in their namespace? Therefore the expectation of a fallthrough due to the priority may be confusing.

npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,100 @@
...
+- Change `spec.priority` field to have a pre-defined value for NetworkPolicy, use numbers below for ANP, and numbers above for BANP.
+For example, if we set NetworkPolicy priority to 0, then negative priorities would mean ANP, and positive priorities would mean BANP.
```
bowei, line 50: This seems a bit too clever...

npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,100 @@
...
+
+While we are here, `priority` field name was brought to our attention multiple times as confusing. It should be mostly related to the fact
+that "higher priority" means "lower number", which is counter-intuitive. Suggested alternatives are "order", "precedence", "sequence".
+
+### Resolving differences between ANP and BANP
+
+1. BANP is a singleton, which is also the reason why it doesn't have `spec.priority` field.  
+Make BANP non-singleton and apply `spec.priority` field to it.
+
+2. ANP supports `Pass` action, while BANP doesn't. `Pass` action could be considered a delegation to the NetworkPolicy,
```
bowei, line 62: Should BANP Pass mean go to NP? But isn't BANP only active if there is no NP? It just sounds like Pass == Allow in behavior?

npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,100 @@
...
+  or it could be seen as return from the current policy layer.   
+We have 2 options: allow using `Pass` for BANP, meaning that the rest od BANPs will be skipped or forbid this action for BANP.  
+Unless we find strong reasons to forbid `Pass` for BANP, it makes sense to allow it to make the API more uniform.
+
+3. ANP supports `domainNames` matching for egress rules, while BANP doesn't.  
+The original reasoning from the 
+[FQDN NPEP](https://github.com/kubernetes-sigs/network-policy-api/blob/feae59d056cf87131338a8449d31b85dc1d9790f/npeps/npep-133-fqdn-egress-selector.md?plain=1#L33-L38)
+is
+> Since Kubernetes NetworkPolicy does not have a FQDN selector, adding this
```
bowei, line 71: Although it sounds like the non-extensibility of NP API was not intended, so likely 

1) NPv2 needs to happen
2) NPv2 will have FQDN (why not?)

bowei (commented):
Mostly discussion comments, general direction is +1 from me (less APIs that are almost 100% duplicated the better)

npeps/npep-285-combine-crds.md
npinaeva: Pass generally means "go the next layer", so for ANP that goes to netpol, and for BANP that goes to the default-cluster-policy, which is allow

npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,100 @@
...
+capability to BaselineAdminNetworkPolicy could result in writing baseline
+rules that can't be replicated by an overriding NetworkPolicy. For example,
+if BANP allows traffic to `example.io`, but the namespace admin installs a
+Kubernetes Network Policy, the namespace admin has no way to replicate the
+`example.io` selector using just Kubernetes Network Policies.
+
+DNS Name selector is only [supported](https://github.com/kubernetes-sigs/network-policy-api/blob/feae59d056cf87131338a8449d31b85dc1d9790f/apis/v1alpha1/adminnetworkpolicy_types.go#L282)
+with the `Allow` rule, which means the desired override by the NetworkPolicy could only be `Deny`. 
+Due to the default-deny nature of NetworkPolicy, all required traffic in a namespace should be explicitly allowed, which means
+the override of the BANP `Allow` DNS Name selector will happen automatically.
```
fasaxc, line 81: I don't follow this; couldn't you have a BANP that had multiple rules:
* deny XYZ
* allow to foo.example.com
* deny to ABC
* allow to bar.example.com

npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,100 @@
...
+that "higher priority" means "lower number", which is counter-intuitive. Suggested alternatives are "order", "precedence", "sequence".
+
+### Resolving differences between ANP and BANP
+
+1. BANP is a singleton, which is also the reason why it doesn't have `spec.priority` field.  
+Make BANP non-singleton and apply `spec.priority` field to it.
+
+2. ANP supports `Pass` action, while BANP doesn't. `Pass` action could be considered a delegation to the NetworkPolicy,
+  or it could be seen as return from the current policy layer.   
+We have 2 options: allow using `Pass` for BANP, meaning that the rest od BANPs will be skipped or forbid this action for BANP.
```
fasaxc, line 64: I think, if Pass is allowed, we could present it as there being a "final default-allow tier" that Pass jumps to.  Then we're free to add more tiers after BANP in future and Pass will do the natural thing.

npeps/npep-285-combine-crds.md
npinaeva: Sure, you can. This part tries to address the original concern that a NetworkPolicy can't override `allow to foo.example.com` part of BANP, which it can, because an override will have to be "deny". So even if we ever implement `allow DNSName` peer in network policy, it has nothing to do with the ability to override BANP.

npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,100 @@
...
+
+Story 1: Multiple BANPs
+
+As a cluster admin, I want to apply different BANPs to different sets namespaces, so that I can have different default trust levels.
+For example, "system" namespaces, like kube-system, could have a less strict BANP than "user" namespaces.
+
+Story 2: Single CRD
+
+As a cluster admin, I prefer to have 1 CRD for both ANP and BANP, so that it is easier to manage, understand, and monitor.
+As an API implementation developer, I prefer to have 1 CRD for both ANP and BANP, so that I can reduce the complexity of the code.
```
tssurya, line 40: Is the code really complex? The fields in BANP are a subset of what we have for ANP

npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,100 @@
...
+As an API developer, I prefer to have 1 CRD for both ANP and BANP, so that I can reduce the complexity of the API and maintenance load.
```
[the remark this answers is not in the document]
tssurya: but I do want to note that in the past 3 years we have advertised ANP/BANP so much that ppl now remember the tiering system so there will also be some confusion once we introduce this change for a little while - at least for people who know the current API - but new users will be fine

npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Resolving differences between ANP and BANP
+
+1. BANP is a singleton, which is also the reason why it doesn't have `spec.priority` field.  
+Make BANP non-singleton and apply `spec.priority` field to it.
+
+2. ANP supports `Pass` action, while BANP doesn't. `Pass` action could be considered a delegation to the NetworkPolicy,
+  or it could be seen as return from the current policy layer.   
+We have 2 options: allow using `Pass` for BANP, meaning that the rest od BANPs will be skipped or forbid this action for BANP.  
+Unless we find strong reasons to forbid `Pass` for BANP, it makes sense to allow it to make the API more uniform.
```
tssurya, line 65: can we add the use case for wanting pass in BANP so that users know when that needs to be used?

npeps/npep-285-combine-crds.md
danwinship: "Pass" is currently documented as "`skip any remaining ANP rules, and then pass execution to any NetworkPolicies that select the pod`", and we generally describe its semantics as being to delegate from the admin to the end users.

But if we change "Pass" to mean "skip to the next tier", and then we add, say, a "HardTenancy" tier between the top tier and the NetworkPolicy tier, then Passing from the top tier would no longer "skip any remaining ANP rules and pass execution to NetworkPolicy". It would just skip from generic ANP rules to tenancy-specific ANP rules, which may or may not be what the admin wanted. (I'm sure we could come up with examples where it would do the right thing, and examples of where it would do the wrong thing.)

Also, I was thinking about the problem of protecting system namespaces from rogue ANPs that accidentally block access to critical components, and thought that perhaps would could add a "Provider" tier above the "Admin" tier, for distro/infra-provider rules (perhaps with a VAP to block new additions to that tier after cluster install time). But that wouldn't work, because the rule you want there is "for subject pods in system namespaces, Pass to the NetworkPolicy tier". But you wouldn't actually be able to Pass directly from the "Provider" tier to the "NetworkPolicy" tier, and "Pass"ing to the "Admin" tier is exactly what you _don't_ want. (Of course, you could instead just have a preinstalled "tier: Admin" policy with priority 0 and a VAP to block other priority 0 policies, and then "Pass" would work. But this seems like the sort of problem that could apply to other potential future tiers.)

npeps/npep-285-combine-crds.md
tssurya: So I typed this comment during our discussion as to not forget to write it down  on npep.. like what you said.. its being added not becausebof a particular usecasd but because its harmless

npeps/npep-285-combine-crds.md
npinaeva: I have a slightly different understanding of the Pass action. I always imagined that having no ANP rules that match a connection should work exactly the same as `Pass`ing that connection to NP (aka just skip the current tier). In the current model ANP `Pass` can be interpreted as "delegate decision to the namespace owner" aka "I don't care what happens to this connection, you decide", which leads me to believe that this action can only be reasonably used as an exception (deny all, but pass all pod-to-pod)

If we add more tiers, I think allowing to skip the whole tier (e.g. with `Pass` always going to the NP level) makes this system more complicated, because then to make sure your tier rules are applied you need to check that a given connection is not skipped on the previous tier. I am also not sure I agree with the "rogue ANP" as a valid use case, because as a cluster admin I want to have control over my cluster connections, and letting provider override my rules makes this system less secure.

npeps/npep-285-combine-crds.md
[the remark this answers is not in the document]
npinaeva: yeah we can, but I think it makes sense to allow it and have more uniform ANP vs BANP experience

npeps/npep-285-combine-crds.md
npinaeva: maintaining 2 copies of something is always harder and more error-prone. How much more complex - depends on the implementation, but there is no way to implement it with 0 overhead

bowei (approved):
My preference is that we get the big change in (remove BNP etc) and there will be multiple passes to fix up the comments etc. From the NPEP perspective, I don't see a huge amount that we can gain by trying to the refining here in this proposal npep-285

bowei:
/lgtm

As per my comment elsewhere -- I'm ok with moving ahead with the large changes and doing refining in the actual API changes themselves vs trying to do it in this proposal.

npeps/npep-285-combine-crds.md
npinaeva: I had an API maintainer user story separately, because 2 CRDs add overhead both for API maintainers and for API implementation devs. While we can ignore our "own" API maintainers problems, making implementors' lives easier sounds like a valid user story to me

npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,140 @@
...
+Story 2: Single CRD
+
+As a cluster admin, I prefer to have 1 CRD for both ANP and BANP, so that it is easier to manage, understand, and monitor.
+As an API implementation developer, I prefer to have 1 CRD for both ANP and BANP, so that I can reduce the complexity of the code.
+
+## API
+
+### Combining ANP and BANP
+
+To combine ANP and BANP, we need to have a way to specify policy layer. Currently, we have the following options:
```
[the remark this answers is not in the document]
npinaeva: do you mean something like integer `tier` values?

bowei:
Repeating my 

/lgtm

My preference for this is to iterate quickly on smaller changes rather than holding a review out for the perfect text. It also allows more than one person to take up the edits.

npeps/npep-285-combine-crds.md
```diff
@@ -0,0 +1,619 @@
...
+	Conditions []metav1.Condition `json:"conditions" patchStrategy:"merge" patchMergeKey:"type"`
+}
+
+// ClusterNetworkPolicySpec defines the desired state of ClusterNetworkPolicy.
+type ClusterNetworkPolicySpec struct {
+	// Tier is used as a first layer of rule prioritization.
+	// ClusterNetworkPolicy has 2 tiers: admin and baseline.
+	// admin tier is used to set strict security rules that can't be overridden.
+	// baseline tier is used to set the default network policy different from "allow all", that can be overridden by NetworkPolicy.
+	// While tiers can only be set on the ClusterNetworkPolicy, they ara some virtual tiers
```
fasaxc, line 170:
```suggestion
	// While ClusterNetworkPolicy can be created in the admin and baseline tiers, other types of policy occupy their own tiers:
```

PR #290, opened 2025-03-24
update to golangci-lint 1.64.7

danwinship:
golangci-lint seems to be failing on `main` for some reason (eg, [here](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_network-policy-api/252/pull-network-policy-api-verify/1904204142750994432), from #252).

Updating to the latest release fixes the problem. :man_shrugging: 

Also, we were previously accidentally committing the golangci-lint binary to github, so fix that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .idea/
 .vscode/
+bin/
 vendor/
 site/
 .DS_Store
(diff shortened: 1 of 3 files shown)


aojea:
/lgtm

bump on golang version? that use to be a problem with golangci-lint

Issue #187, opened 2024-01-09
[ENHANCEMENT] More protocols support 


npinaeva:
**Is your enhancement request related to a problem? Please describe.**
Currently only TCP, UDP, and SCTP protocols are supported, but there are more protocols (ICMP, ICMPv6 are the most popular requests) that may be useful. An example use case is "I want to only allow ICMP connections to implement health monitoring and deny everything else."

**Describe the solution you'd like**
To potentially implement it in the future, we may need to re-consider `AdminNetworkPolicyPort` https://github.com/kubernetes-sigs/network-policy-api/blob/main/apis/v1alpha1/shared_types.go#L52 design, which puts protocol inside the port definition, while some protocols don't have ports, it may be difficult to expand.

**Describe alternatives you've considered**
We could add an extra `protocols` field at the same level as `ports` https://github.com/kubernetes-sigs/network-policy-api/blob/main/apis/v1alpha1/adminnetworkpolicy_types.go#L151, but that will be confusing.

sftim:
As a user, someone may want this extra support to be available for people who only have namespace-level write access.

danwinship:
Making "protocols" top-level instead of "ports" seems a little less convenient/understandable for the common case.

But while we're thinking about how to rearrange things, I've always found `ports` really weird in NP because its behavior is asymmetric; in an ingress rule, `ports` effectively becomes part of the subject, while in an `egress` rule, it becomes part of the peer. Especially, when named ports are used, the `ports` section can affect which pods are part of the subject/peer set (because pods without any matching named port are considered non-matching for the rule).

I don't think there's any easy way to fix that as long as a single policy can have all of (a) a subject, (b) ingress peers, (c) egress peers. If we said that policies had to be either ingress-only or egress-only, then we could move the ports to be part of the subject section in ingress policies, and leave them with the peers in egress policies. But that might be annoying.

[...]

PR #297, opened 2025-04-22
npep-187: More protocols support

npinaeva:
Issue: https://github.com/kubernetes-sigs/network-policy-api/issues/187

### DIFF
--- a/npeps/npep-187-ports-and-protocols.md
+++ b/npeps/npep-187-ports-and-protocols.md
@@ -1,0 +1,104 @@
+# NPEP-187: More protocols support
+
+* Issue: [#187](https://github.com/kubernetes-sigs/network-policy-api/issues/187)
+* Status: Provisional
+
+## TLDR
+
+Change existing `ports` selector to allow more protocols in the future.
+
+## Goals
+
+Change existing `ports` selector to allow more protocols in the future.
+
+## Non-Goals
+
+Adding new protocols to the API.
+
+## Introduction
+
+Currently only TCP, UDP, and SCTP protocols are supported, but there are more protocols (ICMP, ICMPv6 are the most popular requests) 
+that may be useful.
+
+## User-Stories/Use-Cases
+
+Story 1: Protocols extension
+
+As a cluster admin I want to be able to use protocols other than TCP, UDP, and SCTP in my ANP.
+For example, I want to only allow ICMP connections to implement health monitoring and deny everything else.
+
+## API
+
+As we only support port-based protocols (TCP, UDP, and SCTP), current protocol matching is expressed as
+```yaml
+ports:
+  - portNumber:
+      protocol: TCP
+      port: 80
+  - portRange:
+      protocol: UDP
+      start: 1
+      end: 100
+```
+To enable more protocols support in the future, we need to change the high-level field name from `port` to `protocol` to be more generic.
+
+In the `protocols` filter design we decided to start from the YAML that we want (i.e. YAML that is easy to write and understand).
+We want to make popular filters (TCP and UDN ports) easy and less popular filters (like ICMP and named ports) possible.
+This brings us to the following design:
+
+```yaml
+protocols:
+  - protocol: TCP
+    port: 8080
+  - protocol: UDP
+    portRange: 
+      start: 8080
+      end: 9090 
+  - namedPort: http
+    # protocol: "" # must be empty string: defaulted / omitempty
+```
+
+with the validation like so:
+- empty `protocols` list is not allowed
+- at least 1 field for each `protocol` element must be set
+- if `namedPort` is set, `protocol` must be unset
+- if `protocol` is TCP/UDP/SCTP, `port: int` or `portRange: {start: int, end: int}` must be set, but not at the same time
+
+Using ICMP as a potential example for a pure protocol match (`protocol: ICMP`) or protocol with extra fields (e.g. ICMP type and code),
+a future extension could look like this:
+
+```yaml
+protocols:
+  # option 1:  A little inconsistent with port match as it has an extra `icmp` field instead of flat parameters
+  - protocol: ICMP
+    icmp:
+      type: 7
+      code: 3
+  # option 2: Flat parameters, may conflict with the future protocol that also have type or code
+  - protocol: ICMP
+    type: 7
+    code: 3
+  # option 3: Same as option 1, but without a discriminator (protocol)
+  - icmp:
+      type: 7
+      code: 3
+```
+
+## Conformance Details
+
+(This section describes the names to be used for the feature or
+features in conformance tests and profiles.
+
+These should be `CamelCase` names that specify the feature as
+precisely as possible, and are particularly important for
+Extended features, since they may be surfaced to users.)
+
+## Alternatives
+
+(List other design alternatives and why we did not go in that
+direction)
+
+## References
+
+(Add any additional document links. Again, we should try to avoid
+too much content not in version control to avoid broken links)


bowei:
High level looks ok to me plus aligns with what was discussed.

/lgtm

bowei:
How do we make progress on this?

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,222 @@
...
+    - portNumber: 8080
+    - portRange: 
+        start: 8080
+        end: 9090
+  udp:
+    - porNumber: 8080
+    - portNumber: 9090
+  namedPorts:
+    - http
+    - monitoring
```
fasaxc: Feels like a type mismatch that namedPorts is a peer of tcp; how about `any` for this case; it makes it explicit that named ports bring their own protocol with them:
```suggestion
  any:
  - namedPort: http
  - namedPort: monitoring
```
also opens up options like
```
  any:
  - all: true 
```
to explicitly match absolutely anything

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,222 @@
...
+    - type: 7
+      code: 3
+```
+
+In this case it makes more sense to also flatten the `port` match, because for TCP/UDP/SCTP protocols, port is the only option and is always required:
+
+```yaml
+protocols:
+  tcp:
+    - portNumber: 8080
```
fasaxc:
```suggestion
    - port: 8080
```
We don't say `portNumberRange`, could we shorten here too?

npeps/npep-187-ports-and-protocols.md
npinaeva: named port is a protocol+port reference, so it kinda makes sense to me to be on the same level as tcp?
I guess I would agree with `any` if we had more cases like named ports in mind (but I do hope there are none :D)

npeps/npep-187-ports-and-protocols.md
npinaeva: huh makes sense

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,245 @@
...
+      port: 80
+  - portRange:
+      protocol: UDP
+      start: 1
+      end: 100
+```
+To enable more protocols support in the future, we need to change the high-level field name from `port` to `protocol` to be more generic.
+
+We have discussed an option to allow protocol-only match, like TCP any port, but it can also be expressed as a port range 
+that covers all ports (1-65535), so we decided not to add a separate option for protocol-only match.
```
npinaeva: @danwinship I think you had some other arguments to why we shouldn't allow it?

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,104 @@
...
+      start: 8080
+      end: 9090 
+  - namedPort: http
+    # protocol: "" # must be empty string: defaulted / omitempty
+```
+
+with the validation like so:
+- empty `protocols` list is not allowed
+- at least 1 field for each `protocol` element must be set
+- if `namedPort` is set, `protocol` must be unset
```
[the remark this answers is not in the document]
danwinship: We should at least account for it in the design.

(Hm... are we going to have "experimental" fields in v1beta1? That seems kind of contradictory...)

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,104 @@
...
+  - protocol: UDP
+    portRange: 
+      start: 8080
+      end: 9090 
+  - namedPort: http
+    # protocol: "" # must be empty string: defaulted / omitempty
+```
+
+with the validation like so:
+- empty `protocols` list is not allowed
```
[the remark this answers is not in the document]
danwinship: This needs to work the same way as whatever we eventually decided for `ports`.

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,245 @@
...
+
+Option 2:
+```yaml
+protocols:
+  icmp:
+    - type: 7
+      code: 3
+```
+
+A likely option is that we will start with the `ICMP: match all` option, as more details matching is not very popular,
```
danwinship: i think a lot of people want to block pings but not `host unreachable`

npeps/npep-187-ports-and-protocols.md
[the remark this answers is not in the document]
npinaeva: it should be a list, minItems=1, so you can't define an empty list and empty list should work the same as unset

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,245 @@
...
+      code: 3
+```
+
+In this case it makes more sense to also flatten the `port` match, because for TCP/UDP/SCTP protocols, port is the only option and is always required:
+
+```yaml
+protocols:
+  tcp:
+    - port: 8080
+    - portRange:
```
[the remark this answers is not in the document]
danwinship: No, I just meant rename `portRange` to `ports` to make it a little simpler, since the `start` and `end` makes it clear it's a "range" anyway. (I guess this is another briefness/clarity trade-off.)

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,245 @@
...
+- if `namedPort` is set, `protocol` must be unset
+- if `protocol` is TCP/UDP/SCTP, `port` must be set
+
+### Option2: Join by protocol
+
+```yaml
+protocols:
+  tcp:
+    - port:
+        number: 8080
```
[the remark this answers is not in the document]
thockin: IMO, defaulting of port is sort of wart, anyway.  It kind of hurts readability to leave it out.

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,271 @@
...
+  - protocol: UDP
+    port:
+      number: 53
+```
+vs
+```yaml
+protocols:
+  tcp:
+    - port: 53
+  udp:
```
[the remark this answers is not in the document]
bowei: Do we consistently consider the old clients issue? I think inconsistent application of this rule doesn't give us much in terms of guarantees? Specifically, we should define what a client should do when they get something invalid. In the case above, this would be an API object that seemingly violates validation rules that is being sent be the server -- which would unexpected...

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,271 @@
...
+
+A likely option is that we will start with the `ICMP: match all` option, as more details matching is not very popular,
+but we also want to leave space for future extensions (adding more fields), so the struct should allow both
+
+Option 1:
+```yaml
+protocols:
+  - protocol: ICMP
+    icmp:
+      matchAll: true
```
bowei: Is `icmp: {}` an option (was discussed on the Slack thread).

npeps/npep-187-ports-and-protocols.md
bowei: I think v6 would be `icmp6`. I'm looking at the codes and type and they don't line up:

* https://en.wikipedia.org/wiki/Internet_Control_Message_Protocol
* https://en.wikipedia.org/wiki/ICMPv6

npeps/npep-187-ports-and-protocols.md
thockin: Using an empty struct to represent something means that you can never ever add another way of selecting.  This is the same problem with old clients. If a new user specifies one of the newer selection modes, and an older implementation reads it and drops that field, it will assume that means "all", which, for Network policy, could mean failing open. 

It's better for "all" to be an explicit thing. 

In an ideal world, you could perhaps make "all" be the default, applied by the server. If nothing else is specified. Then clients are not subject to the empty structure.  That requires defaulting logic to look at multiple Fields, which isn't very well supported right now, unless you hand code the defaulting.

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,271 @@
...
+    - port:
+        range: 
+          start: 8080
+          end: 9090
+  udp:
+    - port:
+        number: 8080
+    - port:
+        number: 9090
+  namedPort:
```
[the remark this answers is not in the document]
danwinship: @thockin note that NetworkPolicy allows you to specify:

```yaml
ports:
- protocol: TCP
  port: dns-udp
```

(which can't be ruled out at validation time)

and we're trying to avoid that here...

npeps/npep-187-ports-and-protocols.md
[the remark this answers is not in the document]
danwinship: > Do we consistently consider the old clients issue?

We try to (eg https://github.com/kubernetes-sigs/network-policy-api/pull/252).

> This matters most in places where we are likely to add more options over time, and where clients really need to know that they don't know something.

Note that (as we eventually decided in the PR above), this does not currently apply to ANP/CNP, because the implementation will be installing the CRD itself, so it should know all of the fields that exist in the CRD (though if they're installing the experimental CRD, they might not _implement_ every field in the CRD, but in that case they are responsible for noticing that one of the fields they don't implement is set).

> How dangerous is it to get that wrong?

Per the PR above, if the goal is to ensure that no traffic that is supposed to be denied gets accepted instead, then:
- If there's something you don't recognize in an `action: Accept` rule, then you must assume that the rule does not match anything (ie, it neither Accepts nor Denies, but allows further rule processing).
- If there's something you don't recognize in an `action: Deny` rule, then you must assume that the rule matches everything (so, "deny all").
- If there's something you don't recognize in an `action: Pass` rule, then theoretically you could split evaluation in two and follow both paths, but we currently recommend you just treat it as "deny all" as in the previous case.

Of course this is still "dangerous" in the "could totally break your cluster by denying too much" sense, just not in the "accidentally allows forbidden traffic" sense.

npeps/npep-187-ports-and-protocols.md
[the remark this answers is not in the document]
thockin: per-port or for all ports?  i.e. "block port=80 AND flag(PSH)" vs ""block port=80 OR flag(PSH)"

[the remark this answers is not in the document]
npinaeva: I am not sure how this strategy should work with more fields, AFAIU every struct have to have a one-of limitation and add all possible fields combinations as separate fields? See Option3 that I added on the enhancement and lmk if I am missing something

npeps/npep-187-ports-and-protocols.md
fasaxc: > i think a lot of people want to block pings but not host unreachable

Not that we need to figure it out now, but ICMP host unreachable is typically a response packet to an outgoing UDP/TCP packet.  It's best handled as part of the UDP/TCP session from a firewall point of view (I think this is what Linux conntrack does, for example).

npeps/npep-187-ports-and-protocols.md
bowei: Yes, I agree! Having an explicit enum type field description is IMO the most explicit and least error prone path.

In general, I think we should bias towards being more consistent and easy to understand with less emphasis on conciseness because most users will be generating their configuration using an (AI) tool.

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,408 @@
...
+
+Our CRD has only 1 implementation per cluster and both are usually installed and updated together by the cluster admin
+(or a CRD itself may be installed by the implementation).
+So the implementation should know all of the fields that exist in the CRD (though if they're installing the experimental CRD, 
+they might not implement every field in the CRD, but in that case they are responsible for noticing that one of the fields 
+they don't implement is set).
+
+We consider a situation where an older implementation is installed with a newer CRD unlikely/misconfiguration.
+The existing convention on how to handle unknown fields in rules is:
+- If there's something you don't recognize in an `action: Accept` rule, then you must assume that the rule does not match anything
```
bowei: Does a generated client even know if there are unmatched fields? This would require using Unstructured Client for the data structure?

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,408 @@
...
+(or a CRD itself may be installed by the implementation).
+So the implementation should know all of the fields that exist in the CRD (though if they're installing the experimental CRD, 
+they might not implement every field in the CRD, but in that case they are responsible for noticing that one of the fields 
+they don't implement is set).
+
+We consider a situation where an older implementation is installed with a newer CRD unlikely/misconfiguration.
+The existing convention on how to handle unknown fields in rules is:
+- If there's something you don't recognize in an `action: Accept` rule, then you must assume that the rule does not match anything 
+(ie, it neither Accepts nor Denies, but allows further rule processing).
+- If there's something you don't recognize in an `action: Deny` rule, then you must assume that the rule matches everything (so, "deny all").
```
bowei: This seems dangerous?

This seems to only be a problem if the newly added field intersects with the existing sections (only example I can think of is tcp: udp: and we add ip that matches across the previous). Turning a filtered DENY into a DENY ALL is an immediate outage for an application.

It's better to treat these situations as your controller simply got delayed control plane updates unless a problem got fixed. If the control plane signal has problems (for example version skew), it's better to have the data plane coast on a previous known good configuration and alert continuously until it is resolved.

npeps/npep-187-ports-and-protocols.md
```diff
@@ -0,0 +1,408 @@
...
+
+```yaml
+protocols
+- tcp:
+    port: 53
+- udp:
+    port: 53
+- gre:
+    somethingHere: true
+```
```
fasaxc: This has the advantage that "don't restrict on protocol" can be represented by the empty list without defining a sigil.

PR #298, opened 2025-04-22
Fix IPv6 CI by bumping Ubuntu image

danwinship:


.github/workflows/conformance.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         # Install
         sudo cp ${TMP_DIR}/kubectl /usr/local/bin/kubectl
         sudo cp ${TMP_DIR}/kind /usr/local/bin/kind
-        sudo chmod +x /usr/local/bin/*
+        sudo chmod +x /usr/local/bin/kubectl /usr/local/bin/kind
```
danwinship, line 53: first test run failed with:

`chmod: cannot operate on dangling symlink '/usr/local/bin/now'`

so apparently the ubuntu-24.04 image has some junk in it? :man_shrugging:

aojea: yeah, I got bitten for that too

PR #306, opened 2025-07-31
ClusterNetworkPolicy: implementation and clients

npinaeva:
This PR introduced the API changes discussed in the previous iteration https://github.com/kubernetes-sigs/network-policy-api/pull/289 as `v1alpha2`.
Conformance tests for the new API should probably be done in a separate PR, but also nice to be tested before this PR is merged to avoid any mistakes in the API.

code generation script as always inspired by gateway-api, https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/update-clientset.sh

Tested locally with https://github.com/kubernetes-sigs/network-policy-api/pull/307 and https://github.com/kubernetes-sigs/kube-network-policies/pull/199, tests seem to pass!


apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,436 @@
...
+	// While ClusterNetworkPolicy can be created in the admin and baseline tiers, other types of policy occupy their own tiers:
+	// NetworkPolicy tier contains all NetworkPolicies in the cluster and
+	// DefaultTier, which is the default cluster policy, allows all pods to communicate with each other.
+	//
+	// Tiers are evaluated top to bottom, in the following order:
+	// 1. Admin tier
+	// 2. NetworkPolicy tier
+	// 3. Baseline tier
+	// 4. Default tier
+	Tier Tier `json:"tier,omitempty"`
```
aojea, line 73: I really don't understand this very well, so the enum only has two tiers by there can be two additional more?

how does it work?

aojea:
just to confirm, the only diff from v1alpha1 to v1alpha2 are

- type AdminNetworkPolicy + BaselineAdminNetworkPolicy --> ClusterNetworkPolicy
- and  this new `Tier` field inside the new Type, the rest is the same?

I want to get this into kube-network-policies

apis/v1alpha2/clusternetworkpolicy_types.go
npinaeva: yeah we may need to figure out a better way to explain this. Some tiers (Admin and Baseline) belong to the ClusterNetworkPolicy, but there are other tiers in the cluster (NetworkPolicy and Default).
It only defines in which order the policies apply, so in that sense nothing has changed: ANP was always applied before NetworkPolicy and BANP was always applied after NetworkPolicy.
The "new" term here is the Default tier, which simply implies the default "allow-all" policy of any k8s cluster, which just means that without any policies all traffic is allowed.
lmk if you have more questions :)

npinaeva:
> just to confirm, the only diff from v1alpha1 to v1alpha2 are
> 
>     * type AdminNetworkPolicy + BaselineAdminNetworkPolicy --> ClusterNetworkPolicy
> 
>     * and  this new `Tier` field inside the new Type, the rest is the same?
> 
> 
> I want to get this into kube-network-policies

Yes that is a good summary, and thank you for the implementation! 
We do have some more changes planned (some re-work on the protocols field) that will come after this one. So could decide to wait for all changes to come it, or implement it in stages

apis/v1alpha2/clusternetworkpolicy_types.go
npinaeva: that is good feedback, let us think about other options for the Default tier name

PROJECT
```diff
@@ -1,24 +0,0 @@
-domain: policy.networking.k8s.io
```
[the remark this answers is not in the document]
npinaeva: it is used by kubebuilder, see https://github.com/kubernetes-sigs/network-policy-api/pull/306/commits/9a065f333e606ccd31bc025ab239ec509f5c7740 comment

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,436 @@
...
+
+	// Priority only applies within one tier and is a value from 0 to 1000.
```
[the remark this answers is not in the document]
npinaeva: I think 80 symbols will do

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,436 @@
...
+
+const (
+	// ClusterNetworkPolicyRuleActionAllow indicates that matching traffic will be
+	// allowed and no further policy evaluation will be done. This is a final decision.
+	ClusterNetworkPolicyRuleActionAllow ClusterNetworkPolicyRuleAction = "Allow"
+	// ClusterNetworkPolicyRuleActionDeny indicates that matching traffic will be
+	// denied and no further policy evaluation will be done. This is a final decision.
+	ClusterNetworkPolicyRuleActionDeny ClusterNetworkPolicyRuleAction = "Deny"
+	// ClusterNetworkPolicyRuleActionPass indicates that matching traffic will
+	// jump to the next tier evaluation. That means that all the rules with lower
```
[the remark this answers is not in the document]
npinaeva: I guess the idea was to use a different word in case jump is easier to understand for some people? Let's discuss in the meeting

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,482 @@
...
+// ClusterNetworkPolicySpec defines the desired state of ClusterNetworkPolicy.
+type ClusterNetworkPolicySpec struct {
+	// Tier is used as the top-level grouping for network policy prioritization.
+	//
+	// Policy tiers are evaluated in the following order:
+	//
+	// * Admin tier
+	// * NetworkPolicy tier
+	// * Baseline tier
+	// * Default tier
```
fasaxc, line 69: Would prefer a different name for this if we reference it in this way. Calico's "default" tier is the tier that we put k8s network policy in.

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+	// namespace.
+	//
+	// Baseline tier is a cluster-wide policy that can be overridden by the
+	// v1.NetworkPolicy. It will only be evaluated for Pods not selected by a
+	// namespaced v1.NetworkPolicy.
+	//
+	// If a given connection wasn't allowed or denied by any of the tiers,
+	// the default kubernetes policy is applied, which says that
+	// all pods can communicate with each other.
+	Tier Tier `json:"tier"`
```
npinaeva, line 87: I think we haven't discussed this explicitly in the NPEP, but I am making both tier and priority fields required, so no default values. The main difference is that BANP will now have to set some priority, while before there was only 1 BANP allowed

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+	Pods *NamespacedPod `json:"pods,omitempty"`
+}
+
+// NamespacedPod allows the user to select a given set of pod(s) in
+// selected namespace(s).
+type NamespacedPod struct {
+	// NamespaceSelector follows standard label selector semantics; if empty,
+	// it selects all Namespaces.
+	// it selects all Namespaces.
+	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector"`
```
danwinship, line 372: doubled line...

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+
+	// PodSelector is used to explicitly select pods within a namespace;
+	// if empty, it selects all Pods.
+	PodSelector metav1.LabelSelector `json:"podSelector"`
+}
+
+// ClusterNetworkPolicyPort describes how to select network ports on pod(s).
```
danwinship, line 379: "how to select destination network ports".

for egress Nodes and Networks rules this describes ports on hosts, not on pods.

(The ANP docs apparently never got updated to describe this correctly.)

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+// Exactly one field must be set.
+// +kubebuilder:validation:MaxProperties=1
+// +kubebuilder:validation:MinProperties=1
+type ClusterNetworkPolicyPort struct {
+	// Port selects a port on a pod(s) based on number.
```
danwinship, line 384: Port selects a destination port based on protocol and port number.

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+	//
+	// +optional
+	PortNumber *Port `json:"portNumber,omitempty"`
+
+	// NamedPort selects a port on a pod(s) based on name.
```
danwinship, line 389: NamedPort selects a destination port on a pod based on the ContainerPort name.
You can't use this in a rule with Nodes or Networks peers, because they do not have named ports.

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+	//
+	// <network-policy-api:experimental>
+	// +optional
+	NamedPort *string `json:"namedPort,omitempty"`
+
+	// PortRange selects a port range on a pod(s) based on provided
+	// start and end values.
```
danwinship, line 396: PortRange selects a destination port range based on protocol and start and end port numbers.

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+	//
+	// +optional
+	PortRange *PortRange `json:"portRange,omitempty"`
+}
```
danwinship, line 400: should probably order that as Port, PortRange, NamedPort rather than Port, NamedPort, PortRange...

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+	// +kubebuilder:validation:Maximum=65535
+	//
+	Start int32 `json:"start"`
+
+	// End defines a network port that is the end of a port range, the End value
+	// must be greater than Start.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	//
+	End int32 `json:"end"`
```
danwinship, line 437: shouldn't we have CEL validation that Start < End?

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+	// the security policy associated with their deployments inside their
+	// namespace.
+	//
+	// Baseline tier is a cluster-wide policy that can be overridden by the
+	// v1.NetworkPolicy. It will only be evaluated for Pods not selected by a
+	// namespaced v1.NetworkPolicy.
+	//
+	// If a given connection wasn't allowed or denied by any of the tiers,
+	// the default kubernetes policy is applied, which says that
+	// all pods can communicate with each other.
```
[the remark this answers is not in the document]
npinaeva: We first wanted to use the concept of "Default" tier, but that was problematic (see last meeting notes), so we agreed to just explain what happens when no policies match. Not sure how important it is in this context to talk about potential default-deny considering it doesn't exist?

apis/v1alpha2/clusternetworkpolicy_types.go
npinaeva: that is up for discussion - my thinking is that considering we are squashing 2 APIs into 1, it could be easy for people to upgrade to the wrong tier if we have any defaults (also keeping in mind that we wanted to have a very explicit API)

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+	// that cannot be overridden in the other tiers.
+	//
+	// NetworkPolicy tier is the tier for the namespaced v1.NetworkPolicy.
+	// These policies are intended for the application developer to describe
+	// the security policy associated with their deployments inside their
+	// namespace.
+	//
+	// Baseline tier is a cluster-wide policy that can be overridden by the
+	// v1.NetworkPolicy. It will only be evaluated for Pods not selected by a
+	// namespaced v1.NetworkPolicy.
```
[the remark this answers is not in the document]
danwinship: (To be clear, I was just pointing out "we need to revisit this text before the first official v1alpha2 release" not "we need to revisit this text in this PR".)

apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -0,0 +1,479 @@
...
+	// will be allowed and no further policy evaluation will be done.
+	// This is a final decision.
+	ClusterNetworkPolicyRuleActionAllow ClusterNetworkPolicyRuleAction = "Allow"
+	// ClusterNetworkPolicyRuleActionDeny indicates that matching traffic
+	// will be denied and no further policy evaluation will be done.
+	// This is a final decision.
+	ClusterNetworkPolicyRuleActionDeny ClusterNetworkPolicyRuleAction = "Deny"
+	// ClusterNetworkPolicyRuleActionPass indicates that matching traffic
+	// will jump to the next tier evaluation. That means that all the rules
+	// with lower priority at the same tier will be ignored,
```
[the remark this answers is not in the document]
danwinship: This is the whole problem where a "lower priority" policy is one with a numerically-higher `priority` value. That's why we generally always say "higher/lower precedence" instead of "higher/lower priority". (eg, the description of pass in the ANP docs says "lower precedence", not "lower priority")

Issue #308, opened 2025-08-19
[ENHANCEMENT] Extending NetworkPolicy for Multi-Cluster support via `networking.k8s.io/multicluster-name` label


aojea:
**Is your enhancement request related to a problem? Please describe.**

There is no standard way to manage network policies across multiple Kubernetes clusters. This leads to fragmentation as users must rely on project specific solutions.

**Describe the solution you'd like**

Without doing any API change , we can extend the standard NetworkPolicy API to support multi-cluster communication by giving a new meaning to the `networking.k8s.io/multicluster-name` label within the podSelector and namespaceSelector.

**Describe alternatives you've considered**

Continuing to use project specific CRDs is not a portable solution. Creating a new MultiClusterNetworkPolicy CRD would add unnecessary complexity to the API.

**Additional context**

This proposal is inspired by existing multi-cluster network policy implementations in Cilium and Calico. The goal is to create a more consistent and portable user experience.

aojea:
/hold

the label is a bad idea, is ambiguous for some cases

PR #309, opened 2025-08-19
NPEP-308 Extended Network Policy API multicluster via `networking.k8s.io/multicluster-name` label

aojea:
Fixes: #308

### DIFF
--- a/npeps/npep-308.md
+++ b/npeps/npep-308.md
@@ -35,7 +35,7 @@ As multi-cluster architectures become more common, there is a growing need for a
 standardized way to manage traffic flow between services in different clusters.
 Today, this is handled by project-specific solutions, such as Cilium's
 `CiliumNetworkPolicy` or Calico's `GlobalNetworkPolicy`. This leads to a
-fragmented ecosystem where security policies are not portable and useßrs are
+fragmented ecosystem where security policies are not portable and users are
 locked into a specific project for multi-cluster functionality.
 
 This proposal aims to address this by introducing a simple, backward-compatible


npeps/npep-308.md
```diff
@@ -0,0 +1,127 @@
+# NPEP-308: Extending NetworkPolicy for Multi-Cluster support via `networking.k8s.io/multicluster-name` label
+
+* Issue: [#308](https://github.com/kubernetes-sigs/network-policy-api/issues/308)
+* Status: Provisional
+
+## TLDR
+
+This proposal introduces a standardized method for defining multi-cluster
+`NetworkPolicies` by extending the existing Kubernetes `NetworkPolicy` API. This
```
bowei, line 9: There has been a lot of discussion whether or not in the multi-cluster context it makes sense to talk about label selectors vs a better aggregation such as identity. There is a real concern that propagating network policy API in its entirety, which at the cluster-scope level can have scalability issues, to the multi-cluster level will make the scalability even worse.  I think it's very important to have that discussion up front -- as once that is clear, the specific API etc would be a more straightforward exercise.

@robscott @LiorLieberman FYI

npeps/npep-308.md
```diff
@@ -0,0 +1,127 @@
...
+  connectivity and policy enforcement between clusters.
+
+## Introduction
+
+The current Kubernetes `NetworkPolicy` API is limited to intra-cluster traffic.
+As multi-cluster architectures become more common, there is a growing need for a
+standardized way to manage traffic flow between services in different clusters.
+Today, this is handled by project-specific solutions, such as Cilium's
+`CiliumNetworkPolicy` or Calico's `GlobalNetworkPolicy`. This leads to a
+fragmented ecosystem where security policies are not portable and useßrs are
```
danwinship, line 38:
```suggestion
fragmented ecosystem where security policies are not portable and users are
```

npeps/npep-308.md
```diff
@@ -0,0 +1,127 @@
...
+* Issue: [#308](https://github.com/kubernetes-sigs/network-policy-api/issues/308)
+* Status: Provisional
+
+## TLDR
+
+This proposal introduces a standardized method for defining multi-cluster
+`NetworkPolicies` by extending the existing Kubernetes `NetworkPolicy` API. This
+is achieved by giving a new meaning to the well-defined label,
+`networking.k8s.io/multicluster-name`, in the `podSelector` and
+`namespaceSelector`, enabling policies to target resources across different
```
danwinship, line 12: Not `podSelector`. Just `namespaceSelector`. If your peer has no `namespaceSelector` then it only selects pods in the same namespace as the policy, so multicluster is irrelevant. If it has both a `namespaceSelector` and a `podSelector`, then the `podSelector` only selects pods in the namespaces selected by the `namespaceSelector`, so either you have a multi-cluster `namespaceSelector`, or you don't.

npeps/npep-308.md
```diff
@@ -0,0 +1,127 @@
...
+`namespaceSelector` of `NetworkPolicy` resources. When this label is present,
+the CNI is responsible for interpreting it as a selector for resources in a
+remote cluster.
+
+**Importantly, the addition of this label will have no impact on CNIs or other
+projects that do not implement this functionality.** If a CNI does not recognize
+the `networking.k8s.io/multicluster-name` label, it will simply be ignored, and
+the `NetworkPolicy` will be enforced as if the label was not present. This
+ensures full backward compatibility and means that this change will not break
+any existing `NetworkPolicy` implementations.
```
danwinship, line 77: It won't be "ignored", it will be interpreted as being a local label. I guess we assume that nobody will _actually_ label namespaces with this label.

npeps/npep-308.md
```diff
@@ -0,0 +1,127 @@
...
+      app: backend
+  ingress:
+  - from:
+    - podSelector:
+        matchLabels:
+          app: frontend
+      namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: default
+          networking.k8s.io/multicluster-name: cluster-2
```
danwinship, line 101: Where do the cluster names come from?

npeps/npep-308.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
```
danwinship, line 103: When we talked about this, we agreed that a policy with no `networking.k8s.io/multicluster-name` selector always only matches packets in the local cluster. Right?

Do local namespaces have an implicit `networking.k8s.io/multicluster-name` label with the name of the local cluster? Or do they have no `networking.k8s.io/multicluster-name` label? More concretely, does:

```yaml
namespaceSelector:
  matchExpressions:
    - key: "networking.k8s.io/multicluster-name"
      operator: Exists
```

match namespaces in _all_ clusters, including the local cluster, or does it only match namespaces in remote clusters?

npeps/npep-308.md
```diff
@@ -0,0 +1,127 @@
...
+  and syntax for defining multi-cluster policies.
+
+* New MultiClusterNetworkPolicy CRD: This would create a new, dedicated CRD for
+  multi-cluster policies. While this would provide a standard, it would also add
+  another API to learn and manage, and it would not be as seamlessly integrated
+  as extending the existing NetworkPolicy API.
+
+The proposed solution of extending the existing NetworkPolicy API is preferred
+because it is the most lightweight, backward-compatible, and user-friendly
+approach.
```
danwinship, line 121: "magic labels" are kind of sus though...

npeps/npep-308.md
aojea: I'm using the alias from the kubeconfig used, checking at cilium it seems you can define them upfront

npeps/npep-308.md
tssurya: yep I also got confused with why proposal had it for podselector level as well

PR #313, opened 2025-08-27
NPEP-311: Best Practices for Multi-Cluster NetworkPolicy in a Flat Network

### LINKED ISSUE #311: [ENHANCEMENT] Documenting Best Practices for Flat Network Multi-Cluster Network Policy
**Is your enhancement request related to a problem? Please describe.**

This is an informational NPEP to document a best common practices for multi-cluster NetworkPolicy in a flat network model, aligning with conventions from SIG-Multicluster. The goal is to provide a reference for users and implementers. This does not propose any API changes or dictate any behaviors on implementations

**Describe the solution you'd like**

Multi cluster is a complex topic since it conflates multiple problems, but some users just may ONLY want to "stretch" their cluster and apply the same standard Network Policies they have in the single cluster to the objects in the "clusterset".

**Describe alternatives you've considered**

Some bad ideas is to embed the notion of multicluster in the API, as it makes the work unfeasible since it requires to deal with the whole multicluster and identity problem, see https://github.com/kubernetes-sigs/network-policy-api/issues/308

**Additional context**

This is already used in production by Calico so is a behavior we know it works, just documenting it for sharing the experience with the community

aojea:
document some of the existing common practices for applying existing network policies across clusters without increasing the scope of the core kubernetes APIs.

Based on the discussions we had in the SIG Network Policy API meeting on Aug 26 2025, and with feedback from @fasaxc of an existing implementation in production, it will be good to get these  covered on a community place,

Fixes: #311

### DIFF
--- a/npeps/npep-311.md
+++ b/npeps/npep-311.md
@@ -102,7 +102,7 @@ ClusterId](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multi
 
 ### Recommendation
 
-Each namespace within a cluster should be labeled with a key that identifies its
+Each namespace within a cluster should be labeled by the cluster administrator with a key that identifies its
 parent cluster. The recommended label is:
 
 `clusterset.k8s.io/cluster-name: <cluster-id>`
--- a/npeps/npep-311.md
+++ b/npeps/npep-311.md
@@ -105,7 +105,7 @@ ClusterId](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multi
 Each namespace within a cluster should be labeled by the cluster administrator with a key that identifies its
 parent cluster. The recommended label is:
 
-`cluster.clusterset.k8s.io/cluster-name: <cluster-id>`
+`cluster.clusterset.k8s.io: <cluster-id>`
 
 The <cluster-id> value should correspond to the unique name of the cluster
 within the ClusterSet (e.g., cluster-a, us-west-2). This practice enables policy
--- a/npeps/npep-311.md
+++ b/npeps/npep-311.md
@@ -1,4 +1,4 @@
-# NPEP-311: NPEP template
+# NPEP-311: Best Practices for Multi-Cluster NetworkPolicy in a Flat Network 
 
 * Issue:
   [#311](https://github.com/kubernetes-sigs/network-policy-api/issues/311)
--- a/npeps/npep-311.md
+++ b/npeps/npep-311.md
@@ -26,6 +26,9 @@ changes to the API.
 
 ## Non-Goals
 
+* This proposal does not support multi-cluster setups where there are overlapping IPs between clusters (or more generally, any multi-cluster setup where traffic isn't freely routed between clusters).
+
+
 * This proposal does not introduce any changes to the Kubernetes Network Policy
   API specification or any extension based on labels and/or annotations.
 
--- a/npeps/npep-311.md
+++ b/npeps/npep-311.md
@@ -1,0 +1,134 @@
+# NPEP-311: NPEP template
+
+* Issue:
+  [#311](https://github.com/kubernetes-sigs/network-policy-api/issues/311)
+* Status: Informational
+
+## TLDR
+
+This NPEP documents a recommended set of practices for applying standard
+NetworkPolicy resources in a multi-cluster, flat network environment. It
+proposes a conventional labeling scheme, aligned with SIG-Multicluster, to
+enable consistent and predictable cross-cluster policy enforcement without
+changes to the API.
+
+## Goals
+
+* To establish and document a common, intuitive operational model for
+  multi-cluster network policy.
+
+* To provide clear, reusable patterns for administrators to secure applications
+  that span multiple clusters.
+
+* To align these practices with the cluster identification conventions
+  established by SIG-Multicluster in
+  [KEP-2149](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid).
+
+## Non-Goals
+
+* This proposal does not introduce any changes to the Kubernetes Network Policy
+  API specification or any extension based on labels and/or annotations.
+
+* This proposal does not mandate any specific CNI implementation or
+  multi-cluster architecture beyond the prerequisite of a flat network.
+
+## Introduction
+
+In a multi-cluster architecture with a flat network, every pod IP is unique and
+directly routable from any other pod, regardless of its origin cluster. This
+topology allows for a simplified security model where NetworkPolicy selectors
+can be evaluated against a global inventory of all pods and namespaces within a
+defined ClusterSet.
+
+The core principles of this model are:
+
+* Policies are Local: A NetworkPolicy resource is applied only to the cluster
+  where it is created. Policies are not replicated, which contains the impact of
+  changes and allows for per-cluster rollout strategies.
+
+* Selectors are Global: The policy engine within each cluster evaluates policy
+  rules against all known pods and namespaces in the entire ClusterSet.
+
+* Administrator-Defined Identity: The ability to differentiate between clusters
+  is important for cross-cluster communication. This model places the
+  responsibility on the cluster administrator to implement a consistent labeling
+  strategy that can be used for identity and policy selection.
+
+This document formalizes these principles as a set of best practices for the
+community.
+
+## User-Stories/Use-Cases
+
+### Story 1: Securing a "Stretched" Application
+
+As a platform operator running an application across multiple clusters,
+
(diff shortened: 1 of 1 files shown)


npeps/npep-311.md
```diff
@@ -0,0 +1,134 @@
...
+
+The central recommendation is to adopt a consistent labeling scheme for cluster
+identification. To effectively manage policies in a multi-cluster environment,
+it is highly recommended to align with the conventions outlined in [KEP-2149:
+ClusterId](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid).
+
+### Recommendation
+
+Each namespace within a cluster should be labeled with a key that identifies its
+parent cluster. The recommended label is:
```
skitt, line 106: What’s meant by “parent cluster”?

npeps/npep-311.md
```diff
@@ -0,0 +1,134 @@
...
+authors to create selectors that precisely target peers from specific clusters.
+
+## Conformance Details
+
+Not applicable, as this is an informational proposal that does not introduce new
+API features.
+
+## Alternatives
+
+The primary alternative is the absence of a documented best practice. This leads
```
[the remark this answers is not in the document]
skitt: In SIG-MC, namespaces aren’t constrained to single clusters, so clusters and namespaces should be considered orthogonal: a clusterset admin familiar with MCS (multicluster services) would expect a multicluster-aware network policy controller to allow traffic across clusters following namespace constraints. (Yes, network administrators tend not to like this.)

End users ask about cross-cluster behaviour regularly, so it would be good to have some clarity; relying on a `clusterSelector` breaks the orthogonality I just mentioned but I can’t think of a better way right now. So `clusterSelector: {}` would select all clusters (applying namespace and pod selectors as appropriate), and as you say, policies with no `clusterSelector` would only apply to the local cluster.

This also avoids surprising administrators: existing network policies only apply to the local cluster (and if there’s a deny-all, no traffic is allowed across the clusterset).

npeps/npep-311.md
```diff
@@ -0,0 +1,134 @@
+# NPEP-311: Best Practices for Multi-Cluster NetworkPolicy in a Flat Network
```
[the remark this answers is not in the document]
aojea: I couldn't find the right name

> ut if we think this use case is important, then I think we should support it with proper API.

that is what I'm trying to figure out if people just need this stretched network policy mode, if it nodes something else , ...

npeps/npep-311.md
```diff
@@ -0,0 +1,134 @@
...
+  multi-cluster network policy.
+
+* To provide clear, reusable patterns for administrators to secure applications
+  that span multiple clusters.
+
+* To align these practices with the cluster identification conventions
+  established by SIG-Multicluster in
+  [KEP-2149](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid).
+
+## Non-Goals
```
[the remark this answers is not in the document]
aojea: > This proposal does not mandate any specific CNI implementation or
  multi-cluster architecture beyond the prerequisite of a flat network.

it is there, no?

npeps/npep-311.md
```diff
@@ -0,0 +1,134 @@
...
+identification. To effectively manage policies in a multi-cluster environment,
+it is highly recommended to align with the conventions outlined in [KEP-2149:
+ClusterId](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid).
+
+### Recommendation
+
+Each namespace within a cluster should be labeled by the cluster administrator with a key that identifies its
+parent cluster. The recommended label is:
+
+`clusterset.k8s.io/cluster-name: <cluster-id>`
```
aojea, line 108:
```suggestion
`cluster.clusterset.k8s.io/cluster-name: <cluster-id>`
```

npeps/npep-311.md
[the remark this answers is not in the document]
aojea: I think this approach based on @fasaxc feedback, builds on namespace sameness.
To sum up Network policy has two parts:
1. The selector that apply the network policy to the Pods in the namespace where the NP exists, I think this has to be local without doubts, not only because of blast radius problems, but also not doing that imply object propagation across clusters, if the NP only exists in ClusterA why the pods with the same namespace and label on clusterB need to apply the same policy?
2. The selectors that apply to the Ingress and Egress traffic from the selected pods, these are the ones that follow the namespace "sameness" and apply across clusters. Technically this works if the local policy engine has the metadata associated to the IP address from the other clusters.

At this point is when the following point Dan is doing raises, do we want to embed the cluster concept on the API or do we follow the "namespace sameness" concept?

If we consider the cluster concept, it looks to me we definitive need a new CRD, we can not make core APIs cluster aware without extending the scope to all objects (we've seen this in multi network how the feature eaily become viral and touches everything). I'm personally do not have time for persue this and after working on some related projects I do not think is feasible and worth the effort.

If we treat this as the existing [Multi Cluster Services ](https://multicluster.sigs.k8s.io/concepts/multicluster-services-api/) based on the "namespace sameness" concept  I think that is valid and backed by the existing Calico implementation and the Multi Cluster Services KEP. If we follow this path, is this something is raising some red flags or are we ok on iterating? if affirmative, do we document it in SIG MC or in this subproject of SIG Network? if we document it, is it a problem if kube-network-policies implements it?

npeps/npep-311.md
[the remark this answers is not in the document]
aojea: yeah, let me add it

npeps/npep-311.md
aojea: I opened this PR  for exploration, and @skitt you are far expert than anybody here, so if you have ideas I'm all ears , I will take a look at https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid ,

danwinship: The idea is that the NetworkPolicies can refer to namespaces in other clusters via this label. So you can say "allow from namespace foo (in this cluster)", or "allow from namespace foo in cluster bar" or "allow from namespace foo in all clusters", etc.

danwinship: (Though it sounds like maybe that's inconsistent with other multi-cluster APIs)

aojea: because of the "namespace sameness" thing?

skitt: @aojea I understand perfectly that this is about having a discussion :wink:

Thinking about this more, I’m not convince namespace sameness should limit what network policies can do — the way I see it, network policies apply on top of networking defaults, and exist to provide more control over those. If the multicluster default (in the SIG-MC worldview anyway) is that namespaces are the same across a clusterset, that doesn’t prevent network policies from clamping down on what’s possible there. In the same way that network policies nowadays allow administrators to say “I only want traffic between this and this”, it makes sense to me to consider that policies extended to multiples clusters should allow administrators to say “I only want traffic between this cluster and this other cluster”.

That being said, @danwinship I don’t understand your comment about using the label to refer to namespaces in other clusters. Does that mean that the idea here is to create namespaces in a local cluster to represent namespaces in another cluster, with the label indicating that the namespaces are for remote clusters and not the local one? That would indeed be inconsistent with other multicluster APIs where multicluster connectivier (or at least discovery) depends on having the same namespace in multiple clusters…

Put another way, given the practices described here, what would the CRs look like to create a network policy allowing a pod in a local cluster to send traffic to all pods in a given namespace in a remote cluster?

danwinship: > what would the CRs look like to create a network policy allowing a pod in a local cluster to send traffic to all pods in a given namespace in a remote cluster?

So keep in mind that NetworkPolicy is "single-sided": you can say "Pod X is allowed to send traffic to Pod Y", but that doesn't automatically imply that Pod Y will accept packets from Pod X. It just means that if Pod X was "deny-all-egress" before, then it becomes "deny-all-egress-except-egress-to-Pod-Y" now.

So the policy you're asking about is basically just what I gave above, except with `ingress` and `from` changed to `egress` and `to`. But if the remote namespace has its own NetworkPolicies, then it might not accept the packets.

In a single-cluster, with AdminNetworkPolicy, you can write "double-sided" policies that actually say "Pod X can send traffic to Pod Y", which implies both that egress from Pod X to Pod Y and ingress to Pod Y from Pod X are allowed. However, in a multi-cluster context, AdminNetworkPolicy would have the same problem plain NP does with cross-cluster policies. Each AdminNetworkPolicy would only have power over traffic in its own cluster, so an ANP in one cluster would not be able to override policies in a remote cluster.

PR #319, opened 2025-09-11
API: reduce priority range from 1000 to 500

squeed:
This seemingly-minor change means that a full policy tier can be packed into a 16-bit integer (because 500 * 100 < 2^16). As written, the spec could allow up to 100,000 rules per tier, which unfortunately needs 17 bits at a worst case.

### DIFF
--- a/apis/v1alpha2/clusternetworkpolicy_types.go
+++ b/apis/v1alpha2/clusternetworkpolicy_types.go
@@ -80,7 +80,7 @@ type ClusterNetworkPolicySpec struct {
 	// all pods can communicate with each other.
 	Tier Tier `json:"tier"`
 
-	// Priority is a value from 0 to 1000 indicating the precedence of
+	// Priority is a value from 0 to 500 indicating the precedence of
 	// the policy within its tier. Policies with lower priority values have
 	// higher precedence, and are checked before policies with higher priority
 	// values in the same tier. All Admin tier rules have higher precedence than
@@ -94,7 +94,7 @@ type ClusterNetworkPolicySpec struct {
 	// values in cases where ambiguity would be unacceptable.
 	//
 	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=1000
+	// +kubebuilder:validation:Maximum=500
 	Priority int32 `json:"priority"`
 
 	// Subject defines the pods to which this ClusterNetworkPolicy applies.
--- a/config/crd/experimental/policy.networking.k8s.io_clusternetworkpolicies.yaml
+++ b/config/crd/experimental/policy.networking.k8s.io_clusternetworkpolicies.yaml
@@ -791,7 +791,7 @@ spec:
                 type: array
               priority:
                 description: |-
-                  Priority is a value from 0 to 1000 indicating the precedence of
+                  Priority is a value from 0 to 500 indicating the precedence of
                   the policy within its tier. Policies with lower priority values have
                   higher precedence, and are checked before policies with higher priority
                   values in the same tier. All Admin tier rules have higher precedence than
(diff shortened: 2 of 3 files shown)


aojea:
LGTM

/assign @danwinship @bowei

danwinship:
/hold

So, from my perspective as "someone who has not actually implemented ANP myself, and is not currently responsible for maintaining an ANP implementation"...

This is wrong. The numerical limits are not hard API guarantees that implementations can be designed around. The goal is "ensure that the total number of rules is bounded", _not_ "ensure that the total number of rules has a _specific_ bound".

That said, ovn-kubernetes also currently implements prioritization this way (by multiplying "max policies" times "max rules per policy") and in fact, it forces `priority` values to be from 0-99 rather than 0-999, because it needs to fit everything into an even smaller numeric range than you do (because it needs to share the range of OVN ACL priorities with other non-ANP rules). (FTR, ovn-k's restriction of `priority` to the range 0-99 is absolutely, unambiguously non-conformant, and we need a conformance test saying so.)

So, what is the argument for why 2^16 total rules is reasonable, but 2^14 is too few and 2^17 is too many? Or, why is 2^16 rules _per-tier_ the right tradeoff, as opposed to 2^16 total? (Indeed, ovn-k's implementation actually only allows for 2^14 total, not 2^14 per tier... not sure what they're planning to do with BANP priorities... let alone the possibility of DNP in the future...)

Anyway, if we want say that implementations are allowed to assume that "total number of policies" times "total number of rules per policy" times "total number of tiers" is less than a specific value, and that it is reasonable for implementations to be designed in such a way that they would need a complete rearchitecting and rewrite if we increased that value, then that means we need to decide _now_ what the total maximum number of tiers will be in the future (eg, including "DNP", etc). Because once we hit GA, we _can't_ change validation in a way that would reject previously-valid objects, so we can't lower the range of valid priority levels _or_ the maximum length of the rule arrays.

My impression is that we did not actually intend to say that implementations must support 100,000 rules. Rather, the "1000 policies" and "100 rules per policy" were limits that _independently_ made sense, but we assumed that the 1000*100 space would be filled in sparsely. Thus, even if you needed to assign a distinct numeric priority to each rule, you could do this dynamically within a much smaller range than 1000*100.

Or alternatively, if we _did_ intend to say that implementations must support 100,000 rules (and no more), then we should just have _that_ be the requirement, and not try to enforce this in terms of a specific division of max policies / max rules-per-policy. (And thus, say that implementations can assume it's possible to assign a unique priority from 0 to 99,999 to each rule, but that you can't do that _just_ by multiplying.)

danwinship:
OK, consensus in yesterday's meeting was that we expect the "rule space" to be very sparse (that is, we assume that most clusters will not use all 1000 priority values, and most policies will not have 100 rules, and most rules will not have 100 peers, and most `Networks` peers will not have 25 CIDRs, etc).

The `priority` range is API (since if we lowered it in a stable API version it would invalidate some existing policies, while if we raised it in a stable API version, it would change the semantics of some existing policies, since a `priority: 1000` policy would no longer be the final fallback).

However, the `MaxItems` values in the CRDs are _not_ API (or at least, not bidirectionally; we can't lower them in a stable release, but we can raise them). The existing values are arbitrary, and we mostly agreed that they're probably too high, and will be lowering them. But we will reserve the right to raise them in the future, so implementations can't assume they are permanent maximums.

We also generally agreed that the max priority should probably remain a power of 10, and 100 feels too small, so we plan to stick with 1000.

danwinship:
> we expect the "rule space" to be very sparse

(so implementations that need to have distinct priority numbers for every policy/rule/peer/whatever should dynamically map them into whatever range they need, and not try to statically map them)

PR #307, opened 2025-08-13
CNP update tests

npinaeva:
After merging a new v1alpha2 API version, our conformance tests should be updated.
This PR removed ANP+BANP conformance tests, but they can still be used by specifying the previous repo version.
It is not possible anymore to support ANP or BANP only, there is only
one conformance profile now - CNP.
Remove ANP and BANP from the standard and experimental channels.
It also uses the latest kube-network-policy implementation of CNP for our conformance testing.

tracking issue https://github.com/kubernetes-sigs/network-policy-api/issues/285
follow up for https://github.com/kubernetes-sigs/network-policy-api/pull/306

PR #322, opened 2025-09-23
use latest version of kube-network-policies for testing

aojea:
This install kube-network-policies from its main branch, to avoid to have to release a new version everytime we want to test any change on the network policy APIs we add.

kube-network-policy release an independeny binary per NetworkPolicy API to avoid breaking changes on the stable code, so this is safe to iterate, we just need to define which version of the API we are testing in an environment variable.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -13,9 +13,10 @@ on:
 env:
   GO_VERSION: "1.24"
   K8S_VERSION: "v1.33.0"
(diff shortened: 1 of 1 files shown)


aojea:
@npinaeva we just need to merge this, and then you rebase https://github.com/kubernetes-sigs/network-policy-api/pull/307 on top.

kube-network-policies will be already using latest, so we solve the dance of having to release each time we do any change

PR #324, opened 2025-09-24
Fix validation of CIDR fields

danwinship:
re https://github.com/kubernetes/kubernetes/issues/134224; `isCIDR()` allows both "mask-like" CIDRs ("192.168.0.0/24") and "address-like" CIDRs ("192.168.0.5/24"). We only want the former.

(If `isCIDR()` gets changed, then the extra clause here will just be a no-op, but we'll have better backward-compatibility this way.)

### DIFF
--- a/apis/v1alpha1/shared_types.go
+++ b/apis/v1alpha1/shared_types.go
@@ -134,6 +134,6 @@ type AdminNetworkPolicyIngressPeer struct {
 }
 
 // CIDR is an IP address range in CIDR notation (for example, "10.0.0.0/8" or "fd00::/8").
-// +kubebuilder:validation:XValidation:rule="isCIDR(self)",message="Invalid CIDR format provided"
+// +kubebuilder:validation:XValidation:rule="isCIDR(self) && cidr(self) == cidr(self).masked()",message="Invalid CIDR format provided"
 // +kubebuilder:validation:MaxLength=43
 type CIDR string
--- a/apis/v1alpha2/clusternetworkpolicy_types.go
+++ b/apis/v1alpha2/clusternetworkpolicy_types.go
@@ -528,7 +528,7 @@ type PortRange struct {
 // CIDR is an IP address range in CIDR notation
 // (for example, "10.0.0.0/8" or "fd00::/8").
 //
-// +kubebuilder:validation:XValidation:rule="isCIDR(self)",message="Invalid CIDR format provided"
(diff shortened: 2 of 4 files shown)


danwinship:
uh...

`The CustomResourceDefinition "adminnetworkpolicies.policy.networking.k8s.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[egress].items.properties[to].items.properties[networks].items.x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 1.475000x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)`

tssurya:
> @tssurya @npinaeva any ideas about how to deal with that CEL cost error? ^^^

this suggests something is unbounded OR if everything is bounded then yea I remember hitting this error before and I had to play around with the maximum numbers to make it reasonable.
IIRC the problem is that we can have 100 egress rules and each rule can have 100 peers and then each peer could be 25 network cidrs..
the only way to fix this is to reduce this number somewhere..

so see if 20 cidr's or 15 fixes this?

I remember I chose 25 cidrs cause back then as well cost exceeded :)

aojea:
> uh...
> 
> `The CustomResourceDefinition "adminnetworkpolicies.policy.networking.k8s.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[egress].items.properties[to].items.properties[networks].items.x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 1.475000x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)`


> > @tssurya @npinaeva any ideas about how to deal with that CEL cost error? ^^^
> 
> this suggests something is unbounded OR if everything is bounded then yea I remember hitting this error before and I had to play around with the maximum numbers to make it reasonable. IIRC the problem is that we can have 100 egress rules and each rule can have 100 peers and then each peer could be 25 network cidrs.. that's too much? 100x100x25? the only way to fix this is to reduce this number somewhere..
> 
> so see if 20 cidr's or 15 fixes this?
> 
> I remember I chose 25 cidrs cause back then as well cost exceeded for higher numbers.. :)

@jpbetz @cici37 any best practices or guidelines we can follow here?

danwinship:
/hold cancel
/remove-lifecycle rotten

ok, with the changes to limits we can fix this now

npinaeva:
@danwinship the CI is green now, should we merge it?

danwinship:
@npinaeva yes

PR #325, opened 2025-09-24
Update the Makefile help text to include all of the targets

bowei:


danwinship:
/lgtm
/approve

/hold
if you want to take the suggestion first, or just unhold to merge as-is

PR #326, opened 2025-09-24
Update ports to protocols

bowei:
Makes the `ports` clause a more generic `protocols` block to allow for future expansion.

Example

```yaml
apiVersion: policy.networking.k8s.io/v1alpha2
kind: ClusterNetworkPolicy
metadata:
  name: pub-svc-delegate-example
spec:
  tier: Admin
  priority: 20
  subject:
    namespaces: {}
  egress:
  - action: Pass
    to:
    - pods:
        namespaceSelector:
          matchLabels:
            kubernetes.io/metadata.name: bar-ns-1
        podSelector:
          matchLabels:
            app: svc-pub
    protocols:            #<
    - protocol: TCP       #<
      port:               #<
        number: 8080      #<
```

Another example:

```
protocols:
- protocol: TCP
  port:
    range:
      start: 1000
      end: 2000
- protocol: UDP
  port:
    number: 53
```

Ref: https://github.com/kubernetes-sigs/network-policy-api/issues/187

### DIFF
--- a/hack/run-conformance.sh
+++ b/hack/run-conformance.sh
@@ -18,13 +18,58 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-KUBECTL=$(which kubectl)
-KIND=$(which kind)
+usage() {
+  cat <<EOF
+Usage: $0 [commands]
+
+Commands:
+
+  enable_ipv4_and_ipv6_forwarding
+  setup_environment
+  create_cluster
+  install_apis
+  install_np_impl
+  get_cluster_status
+  run_tests
+  export_logs
+
+  all (run all steps)
+
+Example:
+
+  # Run the conformance test locally. Skip some of the github
+  # action steps.
+  ARTIFACTS_DIR=/tmp TMP_DIR=/tmp \\
+    $0 create_cluster \\
+      install_kube_network_policices_from_main \\
+      install_apis \\
+      run_tests
+
+EOF
+}
+
+if [[ "$#" -eq 0 ]]; then
+  usage
+  exit 1
+fi
+
+for arg in "$@"; do
+  case "$arg" in
+    -h|--help)
+      usage
+      exit 0
+      ;;
+  esac
+done
+
+# Use pre-installed kubectl and kind if not running setup_environment.
+KUBECTL=$(which kubectl || echo "not-found")
+KIND=$(which kind || echo "not-found")
 
 echo "[RUN] found KUBECTL=${KUBECTL}"
 echo "[RUN] found KIND=${KIND}"
 
-# These values will be overridden by the github workflow vars.
+# These values can be overriden by external vars.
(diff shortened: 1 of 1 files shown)


apis/v1alpha2/clusternetworkpolicy_types.go
```diff
@@ -316,29 +315,49 @@ type ClusterNetworkPolicyIngressPeer struct {
...
+	// traffic must match. If not specified, this field defaults
+	// to TCP.
+	//
+	// +kubebuilder:default=TCP
+	Protocol corev1.Protocol `json:"protocol,omitempty"`
+
+	// Specific port to match against.
+	//
+	// +optional
+	Port *ClusterNetworkPolicyPort `json:"port,omitempty"`
```
aojea, line 333: is Protocol defaulted if Port is nil? I'm confused on how this will work since Port is optional

npinaeva:
nice! 
One difference form the "perfect yaml" we came up with last time is one extra layer, so you have
```
protocol: TCP
port:
  range:
    start: 1000
    end: 2000
```
instead of 
```
protocol: TCP
portRange:
  start: 1000
  end: 2000
```
is there any special reason for it? I think we wanted to have new embedded fields for non-port-based protocols, like
```
protocol: ICMP
icmp:
  type: 7
  code: 3
```

bowei:
I was looking at some of the other Kubernetes APIs -- and this nesting seems to be more natural, as we are using the prefix instead of nesting. From a future-proofing perspective and validation perspective, it feels like nesting might be the right choice:

Here is what each variant would look like, it doesn't look that bad?

```
protocol: TCP
port:
  number: 80

protocol: UDP
port:
  range:
    start: 1000
    end :1100

protocol: TCP
port:
  name: www

protocol: ICMP
icmp:
  type: 7
  code: 3
```

I'm happy to change it to the other one -- maybe let's quickly discuss this in the meeting to see what others think.

danwinship:
The goal from the KubeCon notes was "Make TCP and UDP easy to write and easy to read. Make ICMP and named ports possible."

So

```yaml
protocols:
  - port: 80
```

is nice because it's very easy for the user. But as we also said in the KubeCon notes, it "[makes] the structs more complicated to make the YAML nicer", and it will make it really hard to understand the `Protocols` struct because there will be a whole bunch of fields, some of which can be specified together and some of which can't. 

I feel like people use ports too much in NetworkPolicies anyway, so I don't really object to making them more annoying to use. :slightly_smiling_face:

danwinship:
> three lines of YAML to specify a single port instead of two.

`protocol: TCP` is the default, so in the common case it's 2-rather-than-1, not 3-rather-than-2

danwinship:
the notes from kubecon say

> protocol defaults to TCP, except with named ports, where it defaults to “”

which I guess addresses my comment above about named ports too... (but the validation and defaulting in this patch need to be updated if that's what we're going with)

PR #331, opened 2025-10-09
Adds unit test to test CRDs

bowei:
crdtest tests the current CRDs against example valid and invalid CRs. There
    are two ways to use this:
    
    Each test case takes its name from the name of the resource being
    instantiated.
    
      - pkg/crdtest/valid are valid resource definitions.
      - pkg/crdtest/invalid are invalid resource definitions.
    
    Run as a standard Go-test:
    
         $ go test ./pkg/crdtest
    
    Run in -watch mode for development:
    
         $ go test -c -o crdtest ./pkg/crdtest # make crdtest
         $ ./crdtest -crdDir config/crd/standard -watch pkg/crdtest/valid 2>/tmp/out
    
    will watch changes to the *.yaml files in pkg/crdtest/valid and attempt to
    load them into an API server. Validation errors will be written to stdout.

Example output for `-watch`:

```
$ ./crdtest -crdDir config/crd/standard -watch pkg/crdtest/valid 2>/tmp/out

apiVersion: policy.networking.k8s.io/v1alpha2
kind: ClusterNetworkPolicy
metadata:
  name: cluster-wide-deny
spec:
  tier: Admin
  priority: 0
  subject:
    namespaces:
      matchLabels:
        kubernetes.io/metadata.name: sensitive-ns
  ingress:
    - action: Denyx
      name: select-all-deny-all
      from:
      - pods:
          namespaceSelector:
            matchLabels: {}
          podSelector:
            matchLabels: {}

❌ pkg/crdtest/valid/cluster-wide-deny.yaml: ClusterNetworkPolicy.policy.networking.k8s.io "cluster-wide-deny" is invalid: [spec.ingress[0].action: Unsupported value: "Denyx": supported values: "Accept", "Deny", "Pass", <nil>: Invalid value: null: some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]
---
apiVersion: policy.networking.k8s.io/v1alpha2
kind: ClusterNetworkPolicy
metadata:
  name: cluster-wide-deny
spec:
  tier: Admin
  priority: 0
  subject:
    namespaces:
      matchLabels:
        kubernetes.io/metadata.name: sensitive-ns
  ingress:
    - action: Deny
      name: select-all-deny-all
      from:
      - pods:
          namespaceSelector:
            matchLabels: {}
          podSelector:
            matchLabels: {}

✅ pkg/crdtest/valid/cluster-wide-deny.yaml
```


pkg/crdtest/main_test.go
```diff
@@ -0,0 +1,221 @@
...
+			fmt.Printf("❌ %s: %v\n", path, err)
+		}
+		klog.Errorf("Failed to apply object from %s: %v", path, err)
+	} else {
+		if *prettyWatch {
+			fmt.Printf("✅ %s\n", path)
+		}
+		klog.Infof("Successfully applied object from %s", path)
+	}
+	fmt.Println()
```
rikatz: was this intentional? to leave some line break or something like that?

pkg/crdtest/main_test.go
```diff
@@ -0,0 +1,221 @@
...
+		// controller-runtime repo: https://raw.githubusercontent.com/kubernetes-sigs/controller-tools/HEAD/envtest-releases.yaml
+		// If the envvar is not passed, the latest GA will be used
+		k8sVersion := os.Getenv("K8S_VERSION")
+
+		var paths []string
+		if *crdDir != "" {
+			paths = []string{*crdDir}
+		} else {
+			paths = []string{
+				filepath.Join("..", "..", "config", "crd", "standard"),
```
npinaeva: I think correct path is 
`filepath.Join("config", "crd", "standard")`, then you also don't need to pass `-crdDir config/crd/standard`

pkg/crdtest/main_test.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get restConfig from BuildConfigFromFlags: %v", err))
+	}
+
+	if *watchDir != "" {
+		watchAndTest(*watchDir)
+		os.Exit(0)
+	}
+
+	rc := m.Run()
```
npinaeva: I expected this path to test all yamls from valid and invalid folders and assure they return expected result. We could then add more test cases e.g. to make sure our CEL works?

pkg/crdtest/main_test.go
bowei: You need to do the ../.. so `go test ./pkg/crdtest` will work. The unit tests run with the working directory.

pkg/crdtest/main_test.go
```diff
@@ -0,0 +1,221 @@
...
+	} else {
+		if *prettyWatch {
+			fmt.Printf("✅ %s\n", path)
+		}
+		klog.Infof("Successfully applied object from %s", path)
+	}
+	fmt.Println()
+}
+
+func watchAndTest(watchDir string) {
```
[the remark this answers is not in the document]
bowei: I tried to put extensive documentation in commit comment and the doc.go.
Is there a more discoverable place where you think I should put the docs?

`-watch` is a common flag for running utilities in inotify mode..

* https://www.typescriptlang.org/docs/handbook/compiler-options.html
* https://jestjs.io/docs/cli
* kubectl --watch

also `watch (1)`

pkg/crdtest/main_test.go
bowei: This does do what you described? The invalid cases expect validation errors (see crd_test.go).

The entry point (main_test.go) has to be named so the "go test -c" will add the right entry point.

npinaeva: right I think what confused me is that running
`./crdtest -crdDir config/crd/standard` won't do anything, but I guess that is ok

pkg/crdtest/main_test.go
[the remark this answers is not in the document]
bowei: updated

bowei:
I incorporated all of the comments.

pkg/crdtest/main_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+	globals.k8sClient, err = client.New(restConfig, client.Options{Scheme: scheme})
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get restConfig from BuildConfigFromFlags: %v", err))
+	}
+
+	if *watchDir != "" {
+		watchAndTest(*watchDir)
+		if testEnv != nil {
+			if err := testEnv.Stop(); err != nil {
```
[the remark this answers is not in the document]
bowei: I changed it to just call os.Exit. I don't like creating more dependencies on the variables (this case `rc`). One thing to note is that I don't think this path will be exercised in most cases as the user will quit out directly using ctrl-c

npinaeva:
/lgtm
@rikatz do you have any other comments?

Issue #341, opened 2025-11-06
[BUG] conformance hostNetwork ports


npinaeva:
**What happened**:
We are using hostNetwork pods in our conformance manifests in range 34345-36366.
https://github.com/kubernetes-sigs/network-policy-api/blob/62b446c7e913dfe57eee2d50784000078cd9ff0e/conformance/base/manifests.yaml#L261-L315
This intersects with the default ephemeral port range > 32768.
This can cause an attempt to bind a port that is already used by another process and leads to test failure.

**What you expected to happen**:
Predictable behaviour for conformance tests.

**How to reproduce it (as minimally and precisely as possible)**:
Bind one of the metnioned port on a worker node and run a test

**Anything else we need to know?**:

danwinship:
Ideally we would let the user pass in a range of acceptable host network ports to use, since different clusters are going to have different firewalling, etc. This would require dynamically generating some things that we currently generate statically, but it might not be bad. Nadia was going to think about this.

PR #345, opened 2025-12-04
[conformance] Make host network ports configurable.

npinaeva:
Add HostNetworkPortRangeStart and HostNetworkPortRangeEnd to
the config to let users specify host port range.
Update Apply logic to execute template with the dynamic values.

Inspired by gateway-api (as always :) ) https://github.com/kubernetes-sigs/gateway-api/blob/9948587970dffd2d9734f0c09da17e524a57a17d/conformance/utils/kubernetes/apply.go#L91-L109

I may need to backport this, because current tests are very flaky d/s

Fixes [#341](https://github.com/kubernetes-sigs/network-policy-api/issues/341)

### DIFF
--- a/conformance/base/admin_tier/experimental-egress-selector-rules.yaml
+++ b/conformance/base/admin_tier/experimental-egress-selector-rules.yaml
@@ -14,7 +14,7 @@ spec:
         matchLabels:
           conformance-house: gryffindor
   egress:
-  - name: "allow-egress-to-36363-on-nodes"
+  - name: "allow-egress-to-{{ index .HostNetworkPorts 0 }}-on-nodes"
     action: "Accept"
     to:
     - nodes:
@@ -23,8 +23,8 @@ spec:
     ports:
       - portNumber:
           protocol: TCP
-          port: 36363
-  - name: "pass-egress-to-5353-on-nodes"
+          port: {{ index .HostNetworkPorts 0 }}
+  - name: "pass-egress-to-{{ index .HostNetworkPorts 2 }}-on-nodes"
     action: "Pass"
     to:
     - nodes:
@@ -33,7 +33,7 @@ spec:
     ports:
       - portNumber:
           protocol: UDP
-          port: 34345
+          port: {{ index .HostNetworkPorts 2 }}
   - name: "deny-egress-to-slytherin-and-nodes-and-internet"
     action: "Deny"
     to:
--- a/conformance/base/manifests.yaml
+++ b/conformance/base/manifests.yaml
@@ -261,55 +261,55 @@ spec:
       containers:
         - name: centaur-client
           image: registry.k8s.io/e2e-test-images/agnhost:2.45
-        - name: centaur-36363-tcp
+        - name: centaur-{{ index .HostNetworkPorts 0 }}-tcp
           image: registry.k8s.io/e2e-test-images/agnhost:2.45
-          command: ["/bin/bash", "-c", "/agnhost serve-hostname --tcp --http=false --port 36363"]
+          command: ["/bin/bash", "-c", "/agnhost serve-hostname --tcp --http=false --port {{ index .HostNetworkPorts 0 }}"]
           ports:
-            - containerPort: 36363
+            - containerPort: {{ index .HostNetworkPorts 0 }}
               protocol: TCP
-              name: web-36363
-        - name: centaur-36364-tcp
+              name: web-{{ index .HostNetworkPorts 0 }}
+        - name: centaur-{{ index .HostNetworkPorts 1 }}-tcp
           image: registry.k8s.io/e2e-test-images/agnhost:2.45
-          command: ["/bin/bash", "-c", "/agnhost serve-hostname --tcp --http=false --port 36364"]
+          command: ["/bin/bash", "-c", "/agnhost serve-hostname --tcp --http=false --port {{ index .HostNetworkPorts 1 }}"]
           ports:
-            - containerPort: 36364
+            - containerPort: {{ index .HostNetworkPorts 1 }}
               protocol: TCP
-              name: web-36364
-        - name: centaur-34345-udp
+              name: web-{{ index .HostNetworkPorts 1 }}
+        - name: centaur-{{ index .HostNetworkPorts 2 }}-udp
           image: registry.k8s.io/e2e-test-images/agnhost:2.45
           # using random http port to avoid conflict with processes on the host (there is no way to disable http port on netexec)
-          command: ["/bin/bash", "-c", "/agnhost netexec --http-port 34358 --udp-port 34345 --udp-listen-addresses $(HOST_IP)"]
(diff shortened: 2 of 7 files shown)


danwinship:
hm, any reason you didn't use go templates? like CNO does, eg, https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/self-hosted/004-config.yaml

conformance/tests/baseline-admin-network-policy-experimental-egress-rules.go
```diff
@@ -111,23 +111,23 @@ var CNPBaselineTierEgressNodePeers = suite.ConformanceTest{
...
 		t.Run("Should support a 'deny-egress' rule policy for egress-node-peer", func(t *testing.T) {
 			// harry-potter-1 is our client pod in gryffindor namespace
 			// ensure egress is DENIED to rest of the nodes from gryffindor; egressRule at index1 should take effect
 			success := kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-gryffindor", "harry-potter-1", "udp",
-				serverPod.Status.PodIP, int32(34346), s.TimeoutConfig.RequestTimeout, false)
+				serverPod.Status.PodIP, int32(s.HostNetworkPortRangeStart+4), s.TimeoutConfig.RequestTimeout, false)
 			assert.True(t, success)
 			success = kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-gryffindor", "harry-potter-1", "sctp",
-				serverPod.Status.PodIP, int32(9003), s.TimeoutConfig.RequestTimeout, false)
+				serverPod.Status.PodIP, int32(s.HostNetworkPortRangeStart+6), s.TimeoutConfig.RequestTimeout, false)
```
danwinship: all this "+6" etc is very non-obvious, but I guess the original hard-coded port numbers were non-obvious too...

(could possibly be `s.HostNetworkPorts[6]` to make it match the yaml better though?)

conformance/utils/kubernetes/apply.go
```diff
@@ -22,6 +22,7 @@ import (
 	"embed"
 	"errors"
 	"fmt"
+	"html/template"
```
danwinship: Should be `text/template`. Same API, just not HTML-specific output semantics.

conformance/base/admin_tier/experimental-egress-selector-rules.yaml
```diff
@@ -14,7 +14,7 @@ spec:
         matchLabels:
           conformance-house: gryffindor
   egress:
-  - name: "allow-egress-to-36363-on-nodes"
+  - name: "allow-egress-to-{{ index .Ports 0 }}-on-nodes"
```
danwinship: Maybe rename the files to `.yaml.in`? Not sure if that's normal/expected...

conformance/utils/suite/suite.go
```diff
@@ -55,6 +56,10 @@ type Options struct {
 	Debug           bool
 	BaseManifests   string
 	NamespaceLabels map[string]string
+	// HostNetworkPortRangeStart allows using a custom port range for host-networked pods.
+	// This is useful to avoid port conflicts on different clusters.
+	// The amount of hostNetwork pods required may change, check base/manifests.yaml {{ .Ports }} indices.
```
danwinship: Is this the "official" documentation for that parameter? If so, mention the default value?
Also, it would probably be good to document exactly how many values are needed (and note elsewhere that if we expand the range that you need to update the value here).

conformance/tests/baseline-admin-network-policy-experimental-egress-rules.go
npinaeva: oh wow you can see I was interrupted a 100 times while writing this (sorry for that, this PR is a bit of a mess)
I think I started with the list indexing, but then decided that no one will pass the full list of ports to use and it may be a better API to just pass `HostNetworkPortRangeStart` (also potentially no change will be required when we need to use more ports as long as the end of the range doesn't intersect with anything)
So what I wanted to do is to use `s.HostNetworkPortRangeStart+6` both in code and yamls, but lmk if you think list is better

conformance/base/admin_tier/experimental-egress-selector-rules.yaml
npinaeva: I think this mostly makes sense if we end up generating a proper `yaml` file after applying the values, but that doesn't happen. We also don't use templating in all yamls, so some tests will have to use `yaml.in` as their manifest and others will use `yaml`.
So I figured it will bring more confusion that clarity

conformance/base/admin_tier/experimental-egress-selector-rules.yaml
```diff
@@ -23,7 +23,7 @@ spec:
     ports:
       - portNumber:
           protocol: TCP
-          port: 36363
+          port: {{ index .HostNetworkPorts 0 }}
   - name: "pass-egress-to-5353-on-nodes"
```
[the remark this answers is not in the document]
npinaeva: huh eagle eye.. let me just put the number as everywhere else for now

conformance/base/manifests.yaml
```diff
@@ -261,55 +261,55 @@ spec:
...
+            - containerPort: {{ index .HostNetworkPorts 1 }}
               protocol: TCP
-              name: web-36364
-        - name: centaur-34345-udp
+              name: web-{{ index .HostNetworkPorts 1 }}
+        - name: centaur-{{ index .HostNetworkPorts 2 }}-udp
           image: registry.k8s.io/e2e-test-images/agnhost:2.45
           # using random http port to avoid conflict with processes on the host (there is no way to disable http port on netexec)
-          command: ["/bin/bash", "-c", "/agnhost netexec --http-port 34358 --udp-port 34345 --udp-listen-addresses $(HOST_IP)"]
+          command: ["/bin/bash", "-c", "/agnhost netexec --http-port {{ index .HostNetworkPorts 3 }} --udp-port {{ index .HostNetworkPorts 2 }} --udp-listen-addresses $(HOST_IP)"]
```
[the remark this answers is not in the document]
npinaeva: heh I hope so, but I also tried to double-check the port numbers

conformance/utils/suite/suite.go
```diff
@@ -121,6 +133,26 @@ func New(s Options) *ConformanceTestSuite {
 func (suite *ConformanceTestSuite) Setup(t *testing.T) {
 	suite.Applier.FS = suite.FS
```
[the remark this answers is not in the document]
npinaeva: the whole New vs Setup is a bit strange. I think Setup is mostly supposed to create the manifests, so the initialization of all things should happen in New, the problem is that New doesn't return an error (or gets a `t *testing.T` passed), so there is no way to do anything that can return an error there without changing the interface. We have `NewConformanceProfileTestSuite` which is a wrapper around a pure `New` call, but I can't guarantee that everyone is using that. Maybe changing `New` signature to return an error is ok?
And to move creation of `Applier` to the `Setup` I need to save `NamespaceLabels` from the config to the `ConformanceTestSuite` which is also kinda ugly.
So I am not sure what is the best way to fix this, opinions are welcome :)

PR #353, opened 2026-01-19
Conformance: Improve test reliability

mazdakn:
ClusterNetworkPolicy conformance tests are flaky due to two issues:
- Re-use of contexts in some tests.
- Lack of a retry in connectivity checker.

This PR improve the quality of conformance tests, and removes sources of flakiness. The proposed changes include:
- Replacing many redundant steps like getting or patching `pod` and `cnp` resources with helper functions, which helps with test readability.
- Using a new context for each operation. Some tests re-used context which led to flaky behavior. 
- Adding a retry mechanism to `PokeServer` utility function, as the desired states are expected to happen eventually. This seems to be a known issue and mentioned here: https://github.com/kubernetes-sigs/network-policy-api/issues/108

In the Calico implementation of ClusterNetworkPolicy, by applying these changes conformance tests passes consistently.


npinaeva:
/lgtm
/approve
Nice improvement, thanks!

PR #357, opened 2026-02-09
Update conformance test for new protocol/port schema

TheBeeZee:
Update the conformance test YAML files to match the recent protocol/port changes in the v1alpha2 CRD.


bowei:
I'm aware of this change...

PR #375, opened 2026-04-15
Add namespace selector changes to conformance tests

### LINKED ISSUE #110: Add namespace selector changes to conformance tests
Currently all conformance tests are testing on a namespace level. However we don't have coverage for example, when the labels of a namespace get updated and policy stops matching OR when labels of a namespace are changed and the new ones start matching an existing policy etc. We should add coverage for these.

/area conformance

anthonyrtong:
Test that traffic is denied when namespace labels are updated and to longer match the matchLabels field in cluster network policy.  This commit fixes #110

### DIFF
--- a/conformance/tests/admin-network-policy-standard-egress-sctp-rules.go
+++ b/conformance/tests/admin-network-policy-standard-egress-sctp-rules.go
@@ -52,6 +52,43 @@ var CNPAdminTierEgressSCTP = suite.ConformanceTest{
 				serverPod.Status.PodIP, int32(9005), s.TimeoutConfig, true)
 		})
 
+		t.Run("Should support a 'deny-egress' policy for SCTP protocol on a namespace selector when namespace labels are changed to no longer match", func(t *testing.T) {
+			// This test uses `egress-sctp` admin CNP
+			// harry-potter-0 is our server pod in gryffindor namespace
+			serverPod := kubernetes.GetPod(t, s.Client, "network-policy-conformance-gryffindor", "harry-potter-0", s.TimeoutConfig.GetTimeout)
+			// luna-lovegood-0 is our client pod in ravenclaw namespace
+			// ensure egress is ALLOWED to gryffindor from ravenclaw
+			// egressRule at index0 will take precedence over egressRule at index1; thus ALLOW takes precedence over DENY since rules are ordered
+			kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-ravenclaw", "luna-lovegood-0", "sctp",
+				serverPod.Status.PodIP, int32(9003), s.TimeoutConfig, true)
+			// luna-lovegood-1 is our client pod in ravenclaw namespace
+			kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-ravenclaw", "luna-lovegood-1", "sctp",
+				serverPod.Status.PodIP, int32(9005), s.TimeoutConfig, true)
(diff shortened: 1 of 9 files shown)


conformance/tests/admin-network-policy-standard-egress-sctp-rules.go
```diff
@@ -52,6 +52,43 @@ var CNPAdminTierEgressSCTP = suite.ConformanceTest{
...
+				serverPod.Status.PodIP, int32(9003), s.TimeoutConfig, true)
+			// luna-lovegood-1 is our client pod in ravenclaw namespace
+			kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-ravenclaw", "luna-lovegood-1", "sctp",
+				serverPod.Status.PodIP, int32(9005), s.TimeoutConfig, true)
+
+			cnp := kubernetes.GetClusterNetworkPolicy(t, s.Client, "egress-sctp", s.TimeoutConfig.GetTimeout)
+			mutate := cnp.DeepCopy()
+			// update namespace selector in egressRule at index0 to match "conformance-house: gryffindor" label
+			mutate.Spec.Egress[0].To[0].Namespaces.MatchLabels = map[string]string{"conformance-house": "gryffindor"}
+			kubernetes.PatchClusterNetworkPolicy(t, s.Client, cnp, mutate, s.TimeoutConfig.GetTimeout)
```
npinaeva: I don't think updating the CNP should be a part of the same test, but since CNP in this test is matching namespace on `kubernetes.io/metadata.name` label there is no way to just update the label.
So I suggest instead adding the namespace label update to the tests that are already matching on `conformance-house`, e.g. `standard-egress-inline-cidr-rules`  has this section 
```
    to:
    - pods:
        namespaceSelector:
          matchLabels:
            conformance-house: slytherin
        podSelector:
          matchLabels:
            conformance-house: slytherin
```
and this part of the test checks that slytherin is allowed https://github.com/kubernetes-sigs/network-policy-api/blob/1c9af17664326c62f9eaa60494c761b3c9fe2d48/conformance/tests/admin-network-policy-standard-egress-inline-cidr-rules.go#L66-L75

So you could add some more steps to this test: update `conformance-house` label to the new value, check the traffic is dropped now, but then also it is important to update the label back. We only create namespaces once to speed up CI, so some other tests may rely on the namespace labels and we should make sure to always put it back as it was.
cc @tssurya I am not sure what was the original idea, did we want to update existing tests or create a new test case for label changes specifically?

npinaeva (commented):
@anthonyrtong thank you for the tests! I am sorry it took me so long to review this. The changes are going in the right direction, but we need to find the right place for them

conformance/tests/admin-network-policy-standard-egress-sctp-rules.go
tssurya: @npinaeva I think what you said above makes sense - we can update existing tests to add this extra step - as long as coverage exists for label update I'm fine with editing existing tests than adding new ones

anthonyrtong: Thanks for the review @npinaeva and @tssurya!  So to summarize the action items for this:
- Move this test behavior to `standard-egress-inline-cidr-rules` or some other test that is matching on `conformance-house`.
- Make sure the behavior of this new test updates the `conformance-house` and then reverts it back afterwards.

npinaeva:
looks good! thank you @anthonyrtong!
/approve
/lgtm

PR #376, opened 2026-04-16
chore(docs): add cnp support in calico

mazdakn:
ClusterNetworkPolicy support is introduced in Calico in v3.32 release. Standard fields and some of experimental fields (network peers and named ports) are supported.


caseydavenport (approved):
LGTM (Pending v3.32 to be released end of April)

frozenprocess:
I would change the last part and just say 3.29 and above.

PR #377, opened 2026-04-21
Add a simple dependabot config

npinaeva:
Out github actions haven't been updated for a while, trying to fix failures like https://github.com/kubernetes-sigs/network-policy-api/actions/runs/24527049130/job/71855392642?pr=376

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -34,7 +34,7 @@ jobs:
       IP_FAMILY: ${{ matrix.ipFamily }}
     steps:
     - name: Check out code
-      uses: actions/checkout@v2
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v5
 
     - name: Enable ipv4 and ipv6 forwarding
       run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/conformance.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       IP_FAMILY: ${{ matrix.ipFamily }}
     steps:
     - name: Check out code
-      uses: actions/checkout@v2
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v5
```
tssurya, line 37: what's with the hash? instead of doing "v5" directly?

npinaeva: the failure is `Error: The actions actions/checkout@v2 and actions/upload-artifact@v4 are not allowed in kubernetes-sigs/network-policy-api because all actions must be pinned to a full-length commit SHA.`
so I both updated the versions and switched to using hashes

rikatz: There is a new enforcement on the Kubernetes org now to use hashes on github actions

tssurya: ah I see thanks @rikatz !

tssurya (approved):
/lgtm

PR #383, opened 2026-06-17
Add CEL validation tests for CRDs

### LINKED ISSUE #148: Add Validation unit tests for the APIs
**Describe the issue**:
We need to enhance our verify lanes to include field specific tests.
I am already working on this, opening an issue to track them

/assign @tssurya 
```[tasklist]
### Tasks
- [ ] Add validation for inline CIDR peers
- [ ] Add validation for kubebuilder syntax specific verification
```

TheBeeZee:
Adds test cases under pkg/crdtest/testdata/ to verify CEL validation rules for ClusterNetworkPolicy (v1alpha2):

- Valid/Invalid CIDRs: Verifies IPv4/IPv6 CIDR format validation.
- Port Ranges: Verifies that start < end constraint is enforced (e.g., rejecting 80-80).
- Egress Constraints (Experimental): Verifies mutual exclusion between egress peers (egress.to) and destination named ports.
  - Note: Experimental test cases (named *.experimental.yaml) are automatically skipped by default when running against standard CRDs, as the validation rules are not present.

Fixes: #148

### DIFF
--- a/pkg/crdtest/testdata/invalid/egress-peer-namedport-conflict.experimental.yaml
+++ b/pkg/crdtest/testdata/invalid/egress-peer-namedport-conflict.experimental.yaml
@@ -1,0 +1,19 @@
+# Invalid: destinationNamedPort cannot be used in combination with network (CIDR) egress peers.
+apiVersion: policy.networking.k8s.io/v1alpha2
+kind: ClusterNetworkPolicy
+metadata:
+  name: egress-peer-namedport-conflict
+spec:
+  tier: Admin
+  priority: 10
+  subject:
+    namespaces:
+      matchLabels:
+        kubernetes.io/metadata.name: sensitive-ns
+  egress:
+    - action: Accept
+      to:
+        - networks:
+            - 10.0.0.0/24
+      protocols:
+        - destinationNamedPort: http
--- a/pkg/crdtest/testdata/invalid/port-range-invalid-equal.yaml
+++ b/pkg/crdtest/testdata/invalid/port-range-invalid-equal.yaml
@@ -1,0 +1,24 @@
+# Invalid: The start port must be strictly less than the end port (start == end is invalid).
+apiVersion: policy.networking.k8s.io/v1alpha2
+kind: ClusterNetworkPolicy
+metadata:
+  name: port-range-invalid-equal
+spec:
+  tier: Admin
+  priority: 10
+  subject:
+    namespaces:
+      matchLabels:
+        kubernetes.io/metadata.name: sensitive-ns
+  egress:
+    - action: Accept
+      to:
+        - namespaces:
+            matchLabels:
+              kubernetes.io/metadata.name: other-ns
+      protocols:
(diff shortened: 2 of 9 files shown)


danwinship:
> Regular contributors should [join the org](https://git.k8s.io/community/community-membership.md#member) to skip this step.

(You can get `kubernetes-sigs` membership even if you're not eligible for `kubernetes` membership too; see further down on the same page.)

danwinship:
/ok-to-test
/approve

I'll leave it for someone else to lgtm.

pkg/crdtest/testdata/invalid/port-range-invalid-equal.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: policy.networking.k8s.io/v1alpha2
```
bowei, line 2: Can we add start > end as well?

pkg/crdtest/testdata/invalid/egress-peer-namedport-conflict.experimental.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: egress-peer-namedport-conflict
+spec:
+  tier: Admin
+  priority: 10
+  subject:
+    namespaces:
+      matchLabels:
+        kubernetes.io/metadata.name: sensitive-ns
+  egress:
+    - action: Accept
```
bowei, line 14: It doesn't hurt to put a comment in the YAML why this is invalid.

TheBeeZee: Thanks, added.

pkg/crdtest/testdata/invalid/port-range-invalid-equal.yaml
TheBeeZee: Thanks, done.

npinaeva:
@TheBeeZee were the last changes something like a rebase or the actual change to the code?

TheBeeZee:
I checked why I repushed after the LGTM and it looks like it was purely a rebase.

TheBeeZee:
@bowei Do you want take another look? Thanks!

TheBeeZee:
The failure in pull-network-policy-api-verify happens because network-policy-api asks for go1.25.0 while the Prow image is using (and enforcing) go1.24.0 - go1.24 formats files differently and causes the test to fail.

Forcing go1.25.0 in hack/update-codegen.sh fixes the issue, but is arguably not the right fix.

I think the right fix woud be to update the Prow image to go 1.25 (or newer), but that's in the k8s repo.

npinaeva:
trying to fix it here https://github.com/kubernetes-sigs/network-policy-api/pull/404
hopefully a rebase after it's merged should solve the problem

TheBeeZee:
> trying to fix it here #404 hopefully a rebase after it's merged should solve the problem

Thanks Nadia, test passes now.

PR #384, opened 2026-06-17
Add conformance tests for PortRange

### LINKED ISSUE #104: Add pod selector matches to conformance test suite

TheBeeZee:
Implements new conformance tests to cover TCP, UDP, and SCTP (Ingress/Egress) port-range tests.

Fixes: #104

### DIFF
--- a/conformance/base/admin_tier/standard-port-range-rules-combined.yaml
+++ b/conformance/base/admin_tier/standard-port-range-rules-combined.yaml
@@ -1,0 +1,99 @@
+apiVersion: policy.networking.k8s.io/v1alpha2
+kind: ClusterNetworkPolicy
+metadata:
+  name: port-range-rules
+spec:
+  tier: Admin
+  priority: 9
+  subject:
+    namespaces:
+      matchLabels:
+        kubernetes.io/metadata.name: network-policy-conformance-gryffindor
+  egress:
+  - name: "allow-to-ravenclaw-tcp-port-range"
+    action: "Accept"
+    to:
+    - namespaces:
+        matchLabels:
+          kubernetes.io/metadata.name: network-policy-conformance-ravenclaw
+    protocols:
+      - tcp:
+          destinationPort:
+            range:
+              start: 80
+              end: 8080
+  - name: "allow-to-hufflepuff-udp-port-range"
+    action: "Accept"
+    to:
+    - namespaces:
+        matchLabels:
+          kubernetes.io/metadata.name: network-policy-conformance-hufflepuff
+    protocols:
+      - udp:
+          destinationPort:
+            range:
+              start: 53
+              end: 5353
+  - name: "allow-to-slytherin-sctp-port-range"
+    action: "Accept"
+    to:
+    - namespaces:
+        matchLabels:
+          kubernetes.io/metadata.name: network-policy-conformance-slytherin
+    protocols:
+      - sctp:
+          destinationPort:
+            range:
+              start: 9003
+              end: 9005
+  - name: "deny-all-egress"
+    action: "Deny"
+    to:
+    - namespaces:
+        matchExpressions:
+        - key: conformance-house
+          operator: Exists
+  ingress:
+  - name: "allow-from-ravenclaw-tcp-port-range"
+    action: "Accept"
+    from:
+    - namespaces:
+        matchLabels:
+          kubernetes.io/metadata.name: network-policy-conformance-ravenclaw
+    protocols:
+      - tcp:
+          destinationPort:
+            range:
+              start: 80
+              end: 8080
+  - name: "allow-from-hufflepuff-udp-port-range"
+    action: "Accept"
+    from:
+    - namespaces:
+        matchLabels:
+          kubernetes.io/metadata.name: network-policy-conformance-hufflepuff
+    protocols:
+      - udp:
+          destinationPort:
+            range:
+              start: 53
+              end: 5353
+  - name: "allow-from-slytherin-sctp-port-range"
+    action: "Accept"
+    from:
+    - namespaces:
+        matchLabels:
+          kubernetes.io/metadata.name: network-policy-conformance-slytherin
+    protocols:
+      - sctp:
+          destinationPort:
+            range:
+              start: 9003
+              end: 9005
+  - name: "deny-all-ingress"
+    action: "Deny"
+    from:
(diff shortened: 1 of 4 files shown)


conformance/base/admin_tier/standard-egress-tcp-port-range-rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: ClusterNetworkPolicy
+metadata:
+  name: egress-tcp-port-range
+spec:
+  tier: Admin
+  priority: 3
+  subject:
+    namespaces:
+      matchLabels:
+        kubernetes.io/metadata.name: network-policy-conformance-durmstrang
```
npinaeva: you could use any namespace except for hufflepuff here, and ensure non-overlapping rules

conformance/tests/admin-network-policy-standard-egress-sctp-rules.go
```diff
@@ -66,6 +69,23 @@ var CNPAdminTierEgressSCTP = suite.ConformanceTest{
 				serverPod.Status.PodIP, int32(9005), s.TimeoutConfig, false)
 		})
 
+		t.Run("Should support an 'allow-egress' policy for SCTP protocol at the specified port range", func(t *testing.T) {
```
npinaeva: I think this is better places at the end of the test, with a comment saying that we are now using `standard-egress-sctp-port-range-rules.yaml` . Otherwise we have a sequence of tests some of which apply changes to the same manifest, so it looks confusing

conformance/utils/suite/suite.go
```diff
@@ -164,13 +164,15 @@ func (suite *ConformanceTestSuite) Setup(t *testing.T) {
 			"network-policy-conformance-hufflepuff",
 			"network-policy-conformance-ravenclaw",
 			"network-policy-conformance-forbidden-forrest",
+			"network-policy-conformance-durmstrang",
```
npinaeva: it would be nice to avoid adding more pods to this suite, I think we can implement the same test cases using existing pods only?

conformance/tests/admin-network-policy-standard-gress-rules.go
```diff
@@ -35,7 +35,10 @@ var CNPAdminTierGress = suite.ConformanceTest{
 	Features: []suite.SupportedFeature{
 		suite.SupportClusterNetworkPolicy,
 	},
-	Manifests: []string{"base/admin_tier/standard-gress-rules-combined.yaml"},
+	Manifests: []string{
+		"base/admin_tier/standard-gress-rules-combined.yaml",
+		"base/admin_tier/standard-gress-omitted-namespaceselector.yaml",
```
npinaeva: I think this test case is complicated enough and we should keep a new one separate (can stay in the same file though). On the other hand we know we want to add pod label update tests too, so it may be reasonable to do it together in another PR and keep this one for port ranges

conformance/base/admin_tier/standard-egress-tcp-port-range-rules.yaml
TheBeeZee: Ah yes, of course. I have removed the new namespace/pods and reused the existing ones.

conformance/utils/suite/suite.go
TheBeeZee: Agreed. I have removed it.

conformance/tests/admin-network-policy-standard-egress-sctp-rules.go
TheBeeZee: I have moved all the port range test cases to the end of the file. I have also added test cases for deny and pass for port ranges.

conformance/tests/admin-network-policy-standard-gress-rules.go
TheBeeZee: I have split this into a separate test and move it into a separate commit. I can move it into a separate PR if you wish.

conformance/tests/admin-network-policy-standard-egress-sctp-rules.go
```diff
@@ -142,5 +145,56 @@ var CNPAdminTierEgressSCTP = suite.ConformanceTest{
 			kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-ravenclaw", "luna-lovegood-1", "sctp",
 				serverPod.Status.PodIP, int32(9005), s.TimeoutConfig, true)
 		})
+
+		// The following tests use `standard-egress-sctp-port-range-rules.yaml` manifest.
```
npinaeva: I think we should be careful about the amount of combinations we want to test.
So here you are adding 12 new files and 3*12 new tests just to test port range, I would suggest that we reduce the number of tests while still trying to cover most of the use cases.
I am thinking of creating 2 separate test files `[b]anp-standard-port-range-rules` and 2 corresponding yaml files with ingress and egress policies in the same file. You could then assign each peer namespace a protocol and add Allow rule for each namespace+protocol+port range with one big deny all at the end. This way we skip separate test for each action type, but we get all protocol combinations in one test.

conformance/tests/admin-network-policy-standard-gress-rules.go
npinaeva: yeah i think it just doesn't cover the full podSelector testing, only one very specific test, so it would be better to move it to another PR with more podSelector-based tests both for subject and peer selection

conformance/tests/admin-network-policy-standard-egress-sctp-rules.go
TheBeeZee: I have split this into a separate test just for port range (one for ANP, one for BANP). Each test has two test cases, one for ingress, another for ingress. TCP/UDP/SCTP are tested in sequence in the same test.

conformance/tests/admin-network-policy-standard-gress-rules.go
TheBeeZee: I have removed the patch for ommitted namespace.

npinaeva:
thank you @TheBeeZee !

Issue #305, opened 2025-07-11
clarify connection-based vs packet-based behavior


danwinship:
NetworkPolicy was originally intended to apply only to connection establishment, _not_ to every packet in a connection. However, like so many other things, this was not documented, and there is inconsistency between implementations currently. (For example, I know that kube-network-policies only checks NP on connection establishment, but ovn-kubernetes checks it on every packet.)

Only enforcing on connection establishment is more efficient, which is probably why we intended it to work that way. But there are users who definitely want the opposite semantics ("when I change policies, any existing connections that are no longer in-policy should be dropped"). That doesn't actually necessarily imply checking every packet: you could do it by deleting conntrack records for out-of-policy connections when policies change. (I think that might actually be what OVN does.)

We should probably say:
- ANP should use packet-based semantics.
- NP implementations _may_ use either semantics
- Future DNP or whatever should be consistent with ANP.

?

(cc @bowei re: https://github.com/kubernetes-sigs/network-policy-api/pull/303#discussion_r2201428943)

danwinship:
> I think we should make a larger study of the ecosystem to understand how this is widely implemented across the industry

I was originally in favor of pushing everyone to the "connection-based" behavior, but there are OpenShift customers who would apparently be pretty upset if we _didn't_ have the "out-of-policy connections get broken immediately when policies change" behavior, whereas it seems unlikely that anyone is strongly in favor of the opposite behavior, so "packet-based" seemed easier to standardize on.

> When a policy changes, the implementation can simply "forget" the state of any existing connections that are no longer allowed.

(Yeah, I mentioned that (and I think that's what OVN actually does).)

> I think instead of "ANP should use packet-based semantics", it should be
> 
> "ANP should be stateful and must re-evaluate active connections when policies change."

👍

> This is the classic Stateful vs Stateless debate on firewalling.

Not entirely; NP implementations have to be stateful anyway since there's always an implied "plus the corresponding return traffic" on every rule.

npinaeva:
In the context of protocols extension: if we ever implement ICMP or any other connection-less protocol, we have to use connection-less logic?

danwinship:
I would say that you don't really need to think about it in that case. But yeah, we need to avoid implying that "ANP only applies to connection-based protocols".

danwinship:
/remove-lifecycle

[...]

PR #386, opened 2026-06-18
conformance: verify CNP ingress/egress directional isolation

### LINKED ISSUE #103: Expand `.Spec.Ingress` and `.Spec.Egress` tests for testing side-effects
- [ ] Currently the ingress-rules test focus on testing ingress traffic towards subject, even with these rules in place, we need to add a new test or two to those files to test egress from subject works fine since there are 0 egress rules in those files.
- [ ] Currently the egress-rules test focus on testing egress traffic towards subject, even with these rules in place, we need to add a new test or two to those files to test ingress from subject works fine since there are 0 ingress rules in those files.

vpbh:
#### What type of PR is this?

/kind test

#### What this PR does / why we need it:

The existing `ClusterNetworkPolicy` ingress/egress rule conformance tests only probe
traffic in the *same* direction as the rules under test — the ingress-rule tests probe
traffic toward the subject, the egress-rule tests probe traffic from the subject. They
never verify the API's directional-independence guarantee:

> CNPs with no egress rules do not affect egress traffic
> CNPs with no ingress rules do not affect ingress traffic

So an implementation with a cross-direction side-effect (e.g. one that treats "the
subject is selected by a policy" as "isolate the pod in both directions", the way plain
NetworkPolicy's `policyTypes` works) could pass the entire suite while silently breaking
the subject's traffic in the *unspecified* direction.

This PR closes that gap. For both the Admin and Baseline tiers it adds:
- an **ingress-only** CNP, and verifies the subject's **egress is unaffected**; and
- an **egress-only** CNP, and verifies the subject's **ingress is unaffected**.

Each test first asserts a positive control (the one-directional rule is actually
enforced) so the isolation assertions are meaningful, then probes the opposite
direction to namespaces the policy does not reference.

New tests (all gated on the standard `ClusterNetworkPolicy` feature):
- `CNPAdminTierIngressRulesEgressUnaffected`
- `CNPAdminTierEgressRulesIngressUnaffected`
- `CNPBaselineTierIngressRulesEgressUnaffected`
- `CNPBaselineTierEgressRulesIngressUnaffected`

#### Which issue(s) this PR fixes:

Fixes #103

#### Special notes for your reviewer:

- The isolation probes deliberately target **ravenclaw** and **hufflepuff** —
  namespaces the policy never references — so the connection's reply path cannot be
  affected by the rule under test and confound the result (e.g. a `deny-to-slytherin`
  egress rule could affect the reply path of an inbound connection *from* slytherin on
  a stateless dataplane). Using unreferenced peers isolates the property cleanly.
- I implemented these as **dedicated tests + minimal single-direction manifests**
  rather than appending probes to the existing `*-rules.go` tests, because those tests
  mutate their policy mid-run (`PatchClusterNetworkPolicy` rule-reordering), which would
  make appended probes order-dependent. Happy to fold them into the existing files
  instead if you'd prefer — the probe logic is identical.
- Scope: TCP only, since directional isolation is protocol-independent. Glad to add
  UDP/SCTP variants if you'd like them.
- No API or codegen changes — purely additive conformance tests.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

### DIFF
--- a/conformance/tests/admin-network-policy-standard-egress-tcp-rules.go
+++ b/conformance/tests/admin-network-policy-standard-egress-tcp-rules.go
@@ -65,6 +65,27 @@ var CNPAdminTierEgressTCP = suite.ConformanceTest{
 				serverPod.Status.PodIP, int32(80), s.TimeoutConfig, false)
 		})
 
+		t.Run("Should not affect ingress traffic when only egress rules are specified", func(t *testing.T) {
+			// The `egress-tcp` admin CNP declares no ingress rules, so the subject's ingress
+			// must be unaffected. Keep this step after an observed DENY (which proves the
+			// policy is programmed) and before the rule reorderings (so it probes the
+			// declared manifest). Probe sources are picked so no rule can match the reply
+			// path even on stateless dataplanes; hufflepuff's "deny everything else" would.
+			// harry-potter-0 is our server pod in gryffindor namespace
+			serverPod := kubernetes.GetPod(t, s.Client, "network-policy-conformance-gryffindor", "harry-potter-0", s.TimeoutConfig.GetTimeout)
+			// luna-lovegood-0 is our client pod in ravenclaw namespace
+			// ensure ingress is ALLOWED from ravenclaw to gryffindor; matches no ingress rules since there are none
+			kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-ravenclaw", "luna-lovegood-0", "tcp",
+				serverPod.Status.PodIP, int32(80), s.TimeoutConfig, true)
+			// harry-potter-1 is our server pod in gryffindor namespace
+			serverPod = kubernetes.GetPod(t, s.Client, "network-policy-conformance-gryffindor", "harry-potter-1", s.TimeoutConfig.GetTimeout)
+			// draco-malfoy-0 is our client pod in slytherin namespace
+			// ensure ingress is ALLOWED from slytherin to gryffindor, even at port 80
+			// where the egress rule denies the opposite direction
+			kubernetes.PokeServer(t, s.ClientSet, &s.KubeConfig, "network-policy-conformance-slytherin", "draco-malfoy-0", "tcp",
+				serverPod.Status.PodIP, int32(80), s.TimeoutConfig, true)
+		})
+
 		t.Run("Should support an 'deny-egress' policy for TCP protocol; ensure rule ordering is respected", func(t *testing.T) {
 			// This test uses `egress-tcp` admin CNP
 			// luna-lovegood-1 is our server pod in ravenclaw namespace
--- a/conformance/tests/admin-network-policy-standard-ingress-tcp-rules.go
+++ b/conformance/tests/admin-network-policy-standard-ingress-tcp-rules.go
@@ -65,6 +65,27 @@ var CNPAdminTierIngressTCP = suite.ConformanceTest{
 				serverPod.Status.PodIP, int32(8080), s.TimeoutConfig, false)
 		})
 
+		t.Run("Should not affect egress traffic when only ingress rules are specified", func(t *testing.T) {
+			// The `ingress-tcp` admin CNP declares no egress rules, so the subject's egress
+			// must be unaffected. Keep this step after an observed DENY (which proves the
+			// policy is programmed) and before the rule reorderings (so it probes the
+			// declared manifest). Probe targets are picked so no rule can match the reply
+			// path even on stateless dataplanes; hufflepuff's "deny everything else" would.
+			// luna-lovegood-0 is our server pod in ravenclaw namespace
+			serverPod := kubernetes.GetPod(t, s.Client, "network-policy-conformance-ravenclaw", "luna-lovegood-0", s.TimeoutConfig.GetTimeout)
+			// harry-potter-0 is our client pod in gryffindor namespace
+			// ensure egress is ALLOWED from gryffindor to ravenclaw; matches no egress rules since there are none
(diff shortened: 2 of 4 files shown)


tssurya:
> #### What type of PR is this?
> /kind test
> 
> #### What this PR does / why we need it:
> The existing `ClusterNetworkPolicy` ingress/egress rule conformance tests only probe traffic in the _same_ direction as the rules under test — the ingress-rule tests probe traffic toward the subject, the egress-rule tests probe traffic from the subject. They never verify the API's directional-independence guarantee:
> 
> > CNPs with no egress rules do not affect egress traffic
> > CNPs with no ingress rules do not affect ingress traffic
> 
> So an implementation with a cross-direction side-effect (e.g. one that treats "the subject is selected by a policy" as "isolate the pod in both directions", the way plain NetworkPolicy's `policyTypes` works) could pass the entire suite while silently breaking the subject's traffic in the _unspecified_ direction.
> 
> This PR closes that gap. For both the Admin and Baseline tiers it adds:
> 
> * an **ingress-only** CNP, and verifies the subject's **egress is unaffected**; and
> * an **egress-only** CNP, and verifies the subject's **ingress is unaffected**.
> 

instead of adding new tests for this, can't we reuse what we do here: https://github.com/kubernetes-sigs/network-policy-api/blob/main/conformance/base/admin_tier/standard-ingress-tcp-rules.yaml for instance? or https://github.com/kubernetes-sigs/network-policy-api/blob/main/conformance/base/admin_tier/standard-egress-tcp-rules.yaml ? i.e I think we already have tests that check just ingress OR egress rules.. so I wouldn't duplicate the structure.. I think its a matter of adding "egress traffic check" to that ingress test right as an extra step?

vpbh:
Hi @tssurya , let me know if you like this! :)

tssurya:
> Hi @tssurya , let me know if you like this! :)

I will, was on PTO will take a look

conformance/tests/admin-network-policy-standard-egress-tcp-rules.go
```diff
@@ -65,6 +65,27 @@ var CNPAdminTierEgressTCP = suite.ConformanceTest{
 				serverPod.Status.PodIP, int32(80), s.TimeoutConfig, false)
 		})
 
+		t.Run("Should not affect ingress traffic when only egress rules are specified", func(t *testing.T) {
+			// The `egress-tcp` admin CNP declares no ingress rules, so the subject's ingress
+			// must be unaffected (issue #103). Keep this step after an observed DENY (which
```
tssurya: nit: do we need to link the issue number really? :D I think just mentioning what we are testing suffices - here and in other places

conformance/tests/admin-network-policy-standard-egress-tcp-rules.go
```diff
@@ -65,6 +65,27 @@ var CNPAdminTierEgressTCP = suite.ConformanceTest{
...
+			// proves the policy is programmed) and before the rule reorderings (so it probes
+			// the declared manifest). Probe sources are picked so no rule can match the reply
+			// path even on stateless dataplanes; hufflepuff's "deny everything else" would.
```
tssurya, line 73: hmmmm this is a grt point. Do we have any implementation specifics we impose on stateless v/s stateful directional connections? I guess we don't prescribe this so this looks correct to me that we should consider it for both styles of implementations

tssurya (commented):
thanks @vpbh for adding these scenarios - just one nit pick on the "we can remove the issue number" otherwise things lgtm, if you update that I can merge this one

conformance/tests/admin-network-policy-standard-egress-tcp-rules.go
vpbh: Ack

conformance/tests/admin-network-policy-standard-egress-tcp-rules.go
vpbh: I dug a bit and I believe we don't prescribe any implementation mechanism, but the spec does mention clues:
* the v1alpha2 action docs say *"Accept: Accepts the selected traffic, **including replies to that traffic** and related ICMP traffic"* ([here](https://github.com/kubernetes-sigs/network-policy-api/blob/3d934508d9f867f5665c44507a2c2eb6e218ffbe/apis/v1alpha2/clusternetworkpolicy_types.go#L193-L195), added in #364).
* I found #305, the goal being to clarify connection-based vs packet-based evaluation — that's , still open, with implementations differing today (kube-network-policies evaluates at connection establishment, ovn-kubernetes per packet, per Dan's issue text). It almost got documented in #366, but the policy-removal section was pulled in review: ([*"that's an open question isn't it? #305"*](https://github.com/kubernetes-sigs/network-policy-api/pull/366#discussion_r3422284935)).

This probe sits in that gap: the inbound connection is allowed by *default* (no ingress rules exist), so no Accept rule fires and the reply guarantee doesn't attach. Whether `deny-to-hufflepuff-everything-else` can then eat the replies depends on how #305 resolves — so I picked sources where both readings give the same verdict. If #305 settles on "replies of allowed connections are always protected", a hufflepuff-sourced probe would make a good follow-up test — I can file an issue for it.

vpbh (commented):
Answered the questions. I will submit a patchset removing the issue details.

conformance/tests/admin-network-policy-standard-egress-tcp-rules.go
tssurya: oh good we have #305 still but yea we don't need to block this on that, we can continue the conversation there.. or open new issue whichever works

tssurya (approved):
/lgtm

PR #388, opened 2026-06-28
conformance: add --dry-run flag to skip dataplane probes

### LINKED ISSUE #107: Add an option for a `--dry-run` in our conformance test suite
See https://github.com/kubernetes-sigs/network-policy-api/pull/98#discussion_r1200840744 for details
/area conformance

vpbh:
When set, the suite applies every manifest and runs all test setup/teardown and API-level logic, but PokeServer skips the dataplane connectivity probe and its assertion. This lets conformance run against a cluster that has the CRDs installed but no NetworkPolicy implementation (e.g. a vanilla kind cluster), catching regressions in the tests, manifests, and CRD schemas independently of any implementation.

The flag threads from flags to suite.Options to ConformanceTestSuite and is propagated to the probe helper via kubernetes.SetDryRun() in Setup, so it applies to both the standard and profile entrypoints. A conformance report generated with --dry-run logs a warning that it must not be submitted. Includes a unit test for the dry-run short-circuit.

Fixes #107

### DIFF
--- a/conformance/conformance_profile_test.go
+++ b/conformance/conformance_profile_test.go
@@ -100,8 +100,8 @@ func TestConformanceProfiles(t *testing.T) {
 }
 
 func testConformance(t *testing.T) {
-	t.Logf("Running conformance profile tests with cleanup: %t\n debug: %t\n enable all features: %t \n supported features: [%v]\n exempt features: [%v]\n conformance profiles: [%v]",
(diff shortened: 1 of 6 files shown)


bowei:
Need to fix your commit message.

npinaeva:
We have closed the issue as stale

PR #390, opened 2026-06-30
Clarify hairpinning behavior in the docs

### LINKED ISSUE #343: resolve implementation-defined behavior around hairpin packets in CNP
There is some unspecified/implementation-defined behavior involving hairpin packets in NetworkPolicy, which is currently also implementation-defined in ClusterNetworkPolicy... We should try to make them well-defined.

Consider a Service X with endpoint Pods A and B, which are both fully isolated.

1. **Simple hairpinning**: If pod A connects to its own podIP, does this succeed despite NetworkPolicy? ("Yes"?)
2. **Service hairpinning**: If pod A connects to Service X and the service proxy redirects the packet back to pod A, does this succeed despite NetworkPolicy? ("Yes"?) (EDIT: or "No"?)
3. **Service hairpinning miss**: If pod A connects to Service X and the service proxy redirects the packet to pod *B*, does this succeed despite NetworkPolicy? ("No"?)
4. **Service masquerading**: If, due to hairpinning or other reasons, the service proxy decides to masquerade a packet, should that packet then be matched against rules for its original source IP or the masqueraded IP? ("Uh....")

I think everyone agrees the answer to the first question is "yes" and the third one is "no". This means the answer to the second question must be inconsistent with one of them. I think these days, everyone says that the answer to the second question is "yes" (obeying the general rule that "NetworkPolicy doesn't see Services", and thus the exception for hairpinned packets from the first case applies). It's possible I'm wrong and Cilium does things differently though? (From some quick searches, it looks like Cilium *used to* do things differently, but now does things the same way everyone else does.)

The fourth case seems unrelated, but I believe that Calico and kube-network-policies don't specifically handle the second case, and they only get it "right" because kube-proxy is masquerading the packet to the node IP, and so then it hits the magic "all packets from the local node are allowed" rule for kubelet probes. So it's worth thinking about whether this is a bug or a feature...

npinaeva:
Fixes #343

### DIFF
--- a/site-src/api-overview.md
+++ b/site-src/api-overview.md
@@ -49,11 +49,26 @@ The ClusterNetworkPolicy (CNP) resource will help administrators set cluster-wid
 rules for the cluster, which are evaluated before or after the NetworkPolicies defined by the
 namespace owners.
 
+ClusterNetworkPolicy is logically enforced outside the Pod
+network namespace. This determines how it applies to "hairpin" traffic that a Pod
+sends to itself:
+
+- A Pod connecting to its own Pod IP is not subject to CNP.
+- A Pod connecting to a Service IP that the service proxy load-balances back
+to the same Pod is subject to ClusterNetworkPolicy.
+
+As a result, a Pod reaching itself by its Pod IP and that same Pod reaching itself
+through a Service IP can behave differently, and this is expected.
+Note that this behavior is based on the logical model of the pod network, 
+and CNP must behave this way even if it is implemented within the pod network namespace, 
+or if traffic from a Pod to a Service IP doesn't leave the pod network namespace.
+
 ### Tiers
 
 Tier is used as the top-level grouping for network policy prioritization.
 
 Policy tiers are evaluated in the following order:
+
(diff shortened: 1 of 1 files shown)


site-src/api-overview.md
```diff
@@ -49,11 +49,23 @@ The ClusterNetworkPolicy (CNP) resource will help administrators set cluster-wid
...
+network namespace. This determines how it applies to "hairpin" traffic that a Pod
+sends to itself:
+
+- A Pod connecting to its own Pod IP is not subject to CNP.
+- A Pod connecting to a Service IP that the service proxy load-balances back
+to the same Pod is subject to ClusterNetworkPolicy.
+
+As a result, a Pod reaching itself by its Pod IP and that same Pod reaching itself
+through a Service IP can behave differently, and this is expected.
+
```
danwinship, line 65: add something like:

Note that this behavior is based on the _logical_ model of the pod network, and CNP must behave this way even if it is implemented within the pod network namespace, or if traffic from a Pod to a Service IP doesn't leave the pod network namespace.

PR #391, opened 2026-07-01
apis: fix DomainName validation pattern to reject invalid characters

vpbh:
The DomainName validation pattern used the character class `[a-zA-z0-9]`.

This is an unintentional typo: the same pattern already uses the correct `[a-zA-Z0-9]` in its other character classes. Correct the two occurrences in the v1alpha1 and v1alpha2 DomainName types and regenerate the CRDs.

The newly-rejected characters are not valid in domain names, so no valid DomainName value is affected.

### DIFF
--- a/apis/v1alpha1/adminnetworkpolicy_types.go
+++ b/apis/v1alpha1/adminnetworkpolicy_types.go
@@ -287,7 +287,7 @@ type AdminNetworkPolicyEgressPeer struct {
 //     "latest.blog.kubernetes.io" match, however "kubernetes.io", and
 //     "wikipedia.org" do not.
 //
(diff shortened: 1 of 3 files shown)


vpbh:
👋 @npinaeva I did a git blame and it seems to be a piece of code you wrote. Honestly, I was understanding the code base with Claude's help and this was flagged by Claude. I think this was a typo? Please correct me if I am wrong!

npinaeva:
nice catch!
/ok-to-test

npinaeva:
You need to run `./hack/update-codegen.sh` to make CI green

vpbh:
Done @npinaeva ! :)

PR #385, opened 2026-06-17
Add conformance tests for same priority disjoint CNPs

TheBeeZee:
Adds a conformance test case to verify that the implementation correctly applies multiple ClusterNetworkPolicies in the Admin tier sharing the same priority but targeting different subjects (disjoint).

- Policies: Two CNPs with priority 15 targeting gryffindor and hufflepuff respectively.
- Assertions: Verifies that allowed traffic for both policies succeeds, and unmatched traffic is correctly blocked by their respective deny rules.

Fixes: #223

PR #396, opened 2026-07-28
conformance: complete same-priority disjoint CNP coverage (Admin + Ba…

### LINKED ISSUE #223: [CONFORMANCE] Add Conformance for "same priority ANPs"; disjoint
**What existing test case does this issue concern**:

**What test case(s) would you like to be added**:

1. We need to add conformance tests for same priority admin network policies that are disjoint with their rules such that both should work as expected

**Why this is needed**:

1. See https://github.com/kubernetes-sigs/network-policy-api/issues/216

FYI: We don't plan to tackle tests for the don't work as expected bits

vpbh:
…seline)

PR #385 added CNPAdminTierSamePriority, covering only the Admin tier with two same-priority policies on different subjects. Issue #223 was reopened because the agreed scope (both tiers; both the same subject/different peers and different subjects/same peer configs) was not yet covered. Fill in the missing tests:

- CNPAdminTierSamePrioritySameSubject: two Admin CNPs at the same priority sharing a subject, with disjoint peers in their Deny rules. This is the case from #216 where only one same-priority policy took effect.
- CNPBaselineTierSamePrioritySameSubject: the Baseline-tier sibling, now that Baseline policies have priorities.
- CNPBaselineTierSamePriority: two Baseline CNPs at the same priority with different subjects referencing the same peer. The rules restrict different ports so any cross-application of one policy's rules to the other policy's subject is observable.

In every test the Deny probes run before the allow probes: observing both Deny rules proves both policies are programmed, so the allow probes cannot pass vacuously while programming is still in progress. Reorder the existing CNPAdminTierSamePriority probes the same way.

Fixes #223


npinaeva (commented):
I wasn't sure in which direction your implementation will go, but I suggest to revert ca0e307370ef096943b78fb46a46e943d3ae084e  and then only add 2 standard-same-priority.go tests in banp and anp. I would also like to have both egress and ingress (you can use the same rules in both direction, just need to add similar checks in the opposite direction). Then I would also add both same-subject and same-peer CNP in the same file (where you now have 2 CNPs)
So in the end we would have 2 new yaml files and 2 new test files

PR #402, opened 2026-09-04
crdtest: test CRDs per API channel

HadrienPatte:
`deny-with-ports.yaml` used `destinationNamedPort`, a field gated behind `<network-policy-api:experimental>`, but the test loaded the standard CRDs by default. Structural pruning dropped the unknown field, leaving two empty protocol entries that then failed `MinProperties=1`, so `TestValid` had been failing. The pruning obscured the cause: the error blamed an empty object rather than the field actually written.

Group testdata by the channel whose CRDs it needs, as `testdata/{standard,experimental}/{valid,invalid}`, and start one envtest API server per channel instead of a single one in `TestMain`. Since the experimental CRDs are a superset of the standard ones, the standard fixtures run against both channels, so a field escaping its experimental gate now surfaces as a standard-channel failure instead of being pruned away silently.

Fixes this failure currently on `main`:
```
--- FAIL: TestValid (2.02s)
    --- FAIL: TestValid/deny-with-ports (0.00s)
        crd_test.go:64: Create() = ClusterNetworkPolicy.policy.networking.k8s.io "deny-with-ports" is invalid: [spec.ingress[0].protocols[4]: Invalid value: 0: spec.ingress[0].protocols[4] in body should have at least 1 properties, spec.ingress[0].protocols[5]: Invalid value: 0: spec.ingress[0].protocols[5] in body should have at least 1 properties], want nil
```


npinaeva:
nice catch! I am wondering though if we should simply use experimental channel for crd test to keep it simple (considering all standard fields are also in the experimental channel)?
Also I think we need to add it to our CI, @bowei wdyt?

PR #403, opened 2026-09-04
conformance: derive report API version from installed CRDs

vpbh:
The conformance profiles report hardcoded `networkPolicyV2APIVersion` to `v0.1.2`, so any report generated since then misstates the API version it certifies: the in-tree CRDs are stamped `v0.2.0`.

This review moves the release-bumped annotation values from `pkg/generator` into an importable `pkg/consts`, and resolve the reported version at suite construction time by reading the `policy.networking.k8s.io/bundle-version` annotation from the CRDs installed in the cluster, following the gateway-api precedent. Missing, partial, or inconsistent annotations, or an installed version that does not match the suite's own bundle version, now fail the profiles suite early with a clear error instead of producing a report that misidentifies the certified API version.

Generated CRD output is unchanged.

### DIFF
--- a/conformance/conformance_profile_test.go
+++ b/conformance/conformance_profile_test.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"testing"
 
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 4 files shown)


npinaeva:
I think you need to make linter happy, otherwise lgtm

PR #404, opened 2026-09-11
Fix toolchain and bump k8s + go versions

npinaeva:
Bump deps to k8s 1.36.4, go 1.26.6


danwinship:
Looks reasonable. What is the GOTOOLCHAIN part for? Just to ensure clearer failures if something tries to use the wrong version in the future?

/lgtm
/approve

PR #405, opened 2026-09-14
Define hairpin traffic capability model for socket-layer load balancers

TheBeeZee:
This PR updates the ClusterNetworkPolicy overview documentation around service hairpin traffic. It explicitly exempts implementations using connect-time load balancing (e.g., eBPF socket LB) from enforcement. A new `HairpinServicePolicyEnforcement` feature is added to the conformance profile framework.

Refs #343


site-src/api-overview.md
```diff
@@ -59,9 +59,13 @@ to the same Pod is subject to ClusterNetworkPolicy.
...
-Note that this behavior is based on the logical model of the pod network, 
-and CNP must behave this way even if it is implemented within the pod network namespace, 
-or if traffic from a Pod to a Service IP doesn't leave the pod network namespace.
+Note that this behavior is based on the logical model of the pod network. In general,
+CNP applies to service hairpin traffic because the traffic logically leaves the Pod to reach
+the service proxy. However, **implementations that perform connect-time load balancing**
+(such as eBPF socket-level load balancing where the Service IP is translated during `connect()`
+inside the pod's network namespace) do not see this traffic leave the netns. These implementations
+are **exempt** from enforcing ClusterNetworkPolicy on service hairpin traffic, provided they document
+this implementation gap while remaining conformant with the specification.
```
danwinship: The docs/API contract should not be assuming that there won't be a later eBPF API that allows implementing this "correctly". Also, I don't think we should say "people using pre-existing technology X are allowed to break this rule if it's more convenient, but people using future technology Y are not".

(Also, this note was added to the wrong section before... it would really fit better in the "General Notes" section above rather than "The ClusterNetworkPolicy Resource")

/hold
for further discussion on the issue

TheBeeZee: I have reworded the text to make it technology-agnostic. I have also added a separate commit that implements a conformance profile for implementations that support enforcement of Hairpin traffic. WDYT?

